PR #278, opened 2024-07-23
feat(cli): show help command when no command present

jarviliam:
This commit is a small change to the cli that runs the --help command when no args are present.


abhinav (approved):
Good call. Thanks!
I have a minor adjustment to make: I'll make it exit with a non-zero code, but that's easy to do in a follow up.

PR #282, opened 2024-07-23
doc: Support tagging sections with "released in"

abhinav:
If there's a comment in the form:

    
    

It will be replaced with a tag specifying which version
the feature was released in.
For unreleased, it'll accurately state
that the feature has not been released yet.


abhinav:

- #282 ◀
    - #283

PR #286, opened 2024-07-23
fix: Use 'git var GIT_EDITOR' for PR body

### LINKED ISSUE #274: branch submit: Use 'git var GIT_EDITOR' to pick editor
Right now, we only use `$EDITOR` to pick editor.
Use of `git var GIT_EDITOR` will be better because it handles all of the following cases:

```
       GIT_EDITOR
           Text editor for use by Git commands. The value is meant to be interpreted by
           the shell when it is used. Examples: ~/bin/vi, $SOME_ENVIRONMENT_VARIABLE,
           "C:\Program Files\Vim\gvim.exe" --nofork. The order of preference is the
           $GIT_EDITOR environment variable, then core.editor configuration, then
           $VISUAL, then $EDITOR, and then the default chosen at compile time, which is
           usually vi.
```

As noted above, we'll want to interpret the output via a shell (e.g. `bash`), so if the editor is `FOO=bar myeditor`, it still works.

abhinav:
We were using `$EDITOR` to fill in the PR body.
This is not correct.
We should use the output of 'git var GIT_EDITOR'.
The Git editor (and `$EDITOR`) are allowed to be shell commands,
not just executable names. Handle that correctly too.

Lastly, re-use the same logic for stack edit and downstack edit.

Resolves #274

### DIFF
--- a/.changes/unreleased/Fixed-20240723-070944.yaml
+++ b/.changes/unreleased/Fixed-20240723-070944.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'branch submit: Use the same editor used by Git for commit messages to author the PR body.'
+time: 2024-07-23T07:09:44.524288-07:00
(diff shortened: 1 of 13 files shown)


abhinav:

- #286 ◀

PR #295, opened 2024-07-24
Add Simple Secret Store

### LINKED ISSUE #294: Alternative Secret Store
No matter what auth method I try, I keep getting the following error:
```
FTL gs: The name org.freedesktop.secrets was not provided by any .service files
```

I'm running in a headless docker container, without GNOME or KDE installed. I'm unaware of any headless options for a org.freedesktop.secret service.

It would nice if there was some way to store github tokens/credentials without a secret store (or point in the documentation to a headless secret store if we can find one).

One alternative might be storing the token in a dedicated file in $XDG_DATA_HOME, i.e. maybe `~/.local/share/git-spice/github-token`, with the perms set to 400 so only the user can read it.

klnusbaum:
Certain environments (namely headless linux) don't have a keychain
available. Let's make a simple, JSON file secret store for uses in these
environments.

Fixes #294

### DIFF
--- a/internal/secret/simple.go
+++ b/internal/secret/simple.go
@@ -1,0 +1,77 @@
+package secret
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+)
+
+// SimpleStore is a simple secret store.
+//
+// Secrets are stored in a protected file, in JSON format, in $XDG_DATA_HOME
+type SimpleStore struct {
+	location string
+}
+
+var _ Stash = (*SimpleStore)(nil)
+
+type SimpleData struct {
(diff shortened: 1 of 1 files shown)


klnusbaum:
What I've got here now is just a __very__ rough sketch.

If you think this is a good idea I can:
1. Hook it up to the relevant commands
2. Write tests for it
3. Make it thread safe?

abhinav:
Oops, we had a bit of a race there. I implemented something similar locally. 
I tried to merge it into this PR but GitHub isn't letting me push to this branch, so closing this in favor of #296.

PR #296, opened 2024-07-24
secret: Add Unsafe stash

abhinav:
Adds a stash that stores secrets in plain text in a local JSON file.
For safety and visibility, advertises the location of the stash
the first time it's written to.

PR #297, opened 2024-07-24
secret: Add Fallback stash

abhinav:
Adds a secret storage that will allow a primary and secondary store.
The secondary store is used only if the primary is broken
for non-expected reasons.

Notably, LoadSecret returning ErrNotFound is not an unexpected breakage.


abhinav:

- #296
    - #297 ◀

PR #301, opened 2024-07-24
branch submit: Add -c short form for --fill

klnusbaum:
I use the `--fill` option a lot. Can we get a short flag for it?

### DIFF
--- a/branch_submit.go
+++ b/branch_submit.go
@@ -21,7 +21,7 @@ import (
 // submitOptions defines options that are common to all submit commands.
 type submitOptions struct {
 	DryRun bool `short:"n" help:"Don't actually submit the stack"`
-	Fill   bool `help:"Fill in the change title and body from the commit messages"`
+	Fill   bool `short:"c" help:"Fill in the change title and body from the commit messages"`
 	// TODO: Default to Fill if --no-prompt?
 	Draft     *bool `negatable:"" help:"Whether to mark change requests as drafts"`
 	NoPublish bool  `name:"no-publish" help:"Push branches but don't create change requests"`
--- a/.changes/unreleased/Added-20240724-121944.yaml
+++ b/.changes/unreleased/Added-20240724-121944.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: New short flag, `-c`, for the `--fill` flag in the `gs branch submit` command.
+time: 2024-07-24T12:19:44.56674951-07:00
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and is generated by [Changie](https://github.com/miniscruff/changie).
 ### Added
 - Publish pre-built Linux ARM binaries.
 - {branch, stack, upstack, downstack} submit: Add --force flag. This acts like 'git push --force'.
+- Short flag for gs branch submit --file (added -f)
```
abhinav, line 12: This repo uses [changie](https://changie.dev/) to track unreleased changes.
This file only contains things that have already been released.
(https://github.com/abhinav/git-spice/blob/main/CONTRIBUTING.md#making-contributions.)

branch_submit.go
```diff
@@ -21,7 +21,7 @@ import (
 // submitOptions defines options that are common to all submit commands.
 type submitOptions struct {
 	DryRun bool `short:"n" help:"Don't actually submit the stack"`
-	Fill   bool `help:"Fill in the change title and body from the commit messages"`
+	Fill   bool `short:"f" help:"Fill in the change title and body from the commit messages"`
```
abhinav, line 24: `f` has risk of being confused for `--force`, which currently doesn't have a short flag, but could conceivably get one.

Instead of `f` as short for fill, how does `F` or `C` (as in "use my commits") sound?

abhinav (commented):
This makes sense. Left some comments.
FYI, after adjusting the options, you'll need to run `make generate` to update the CLI reference in the documentation.

branch_submit.go
klnusbaum: I'm fine with `-F`. That said, `--force` seems like the more dangerous of the options, so should that be the one with the capital letter? (i.e. we wanna have a speed bump with `--force`, so we'd want it's short option to be the one that takes a little longer to type).

CHANGELOG.md
klnusbaum: oh woops, let me fix this

branch_submit.go
abhinav: I agree—`--force` should have a speed bump. But:

- Git uses `-f`/`--force`. Matching Git isn't a hard requirement, but it's a matter of convention and expectations.
- `-f` as short for `--force` is reasonably common in other commonly used commands, including `rm`.
- There's already `-f`/`--force` on `branch delete`, so short form here will probably want to match.

I think I'm in favor of dropping the short form `-f` from `branch delete` because a) the operation is dangerous, so the speedbump, and b) that command will prompt you for a force-like behavior anyway, if needed. (I'll do that separately from this PR.)

But I'm still having trouble using `-f` for something not `--force` given how prevalent that short form is for that flag. Looking at `git commit` for inspiration:

- `-C <commit>` takes the new commit message from the given commit
- `-c <commit>` takes the new commit message from the given commit and opens an editor
- `-F <file>` takes the commit message from a file

While we don't take an argument with --fill, any of those three seem much more digestible here.

PR #302, opened 2024-07-24
branch delete: Drop short-form --force flag

abhinav:
In interactive mode, the command will prompt for force-deletion
regardless.
In non-interactive mode, the long-form isn't much of a burden.


abhinav:

- #302 ◀

PR #311, opened 2024-07-28
Check if $EDITOR env var is empty

### LINKED ISSUE #310: gs bs: bail early if $EDITOR variable not set
Was trying this tool out on a new linux box earlier, and noticed that on a machine w/o $EDITOR variable configured, the prompt takes me straight to ask whether the PR should be marked as draft before bailing with an error on running the empty command from open editor action.

Ex)
```
❯ gs bs
Title: Remove glide.yaml
Title: Remove glide.yaml
Body: Press [e] to open  or [enter/tab] to skip
no editor found: please set $EDITOR
Draft: [y/N]
FTL gs: prompt form: no editor found: please set $EDITOR
```

sywhang:
Currently `OpenEditor` does not take into account when the $EDITOR env var is not set - in which case the Command just errors out and continues to ask the next question on whether to mark the PR as draft.

The fix is quite simple - add a case for checking if the env var is set and return tea.Quit if that's the case. 

Fix #310.

### DIFF
--- a/internal/ui/open_editor.go
+++ b/internal/ui/open_editor.go
@@ -194,6 +194,10 @@ func (a *OpenEditor) Update(msg tea.Msg) tea.Cmd {
 
 		case key.Matches(msg, a.KeyMap.Accept):
 			return AcceptField
+
+		case a.Editor.Command == "":
+			a.err = errors.New("please set $EDITOR environment variable")
+			return tea.Quit
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


internal/ui/open_editor.go
```diff
@@ -194,6 +194,10 @@ func (a *OpenEditor) Update(msg tea.Msg) tea.Cmd {
 
 		case key.Matches(msg, a.KeyMap.Accept):
 			return AcceptField
+
+		case a.Editor.Command == "":
+			a.err = errors.New("please set $EDITOR environment variable")
```
abhinav, line 199: Suggestion: 
> please configure a Git core.editor or set the $EDITOR environment variable

abhinav (approved):
This makes sense but I want to confirm the version you saw the issue on, and what the output of “git var GIT_EDITOR” was.

abhinav:
Okay, so #286 should have fixed this, but there's still a bug:
if there's no Git editor, and no editor, this may still result in an empty editor.
This makes sense to me.

PR #315, opened 2024-07-29
doc/go.mod: Don't use 'replace'

abhinav:
Don't need to have a go.mod-dependency on the gs binary.
Can just cd .. and build.


abhinav:

- #315 ◀

PR #323, opened 2024-08-05
refac: Extract shorthand support into a separate package

abhinav:
Extracts support for adding shorthands to the CLI
into a separate package with composable "Source" types
for providing the expansions.
We start with only a built-in source, with rooom for more in the future.

This also ends up testing the shorthand expansion thoroughly.


abhinav:

- #323 ◀

PR #326, opened 2024-08-07
Add a configuration system based on git-config

### LINKED ISSUE #16: Configuration system
Add a configuration system: repo-level and user-local.
Following configurations are probably needed:

- branch name prefix: common prefix for branches created with `branch create`. e.g. `$username/`.

abhinav:
git-spice needs a source of configuration
to support customizing behaviors like `submit` navigation comments,
custom aliases, etc.
After some deliberation, we've decided to use git-config
as the configuration source for git-spice.

This will allow for a familiar configuration experience
with user-, repository-, and worktree-level overrides for options.

This PR does not implement any configuration options yet.
Those will be added in subsequent PRs.

Resolves #16

### DIFF
--- a/internal/git/config.go
+++ b/internal/git/config.go
@@ -1,0 +1,145 @@
+package git
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/charmbracelet/log"
+)
+
+// Config provides access to Git configuration in the current context.
+type Config struct {
+	log  *log.Logger
+	dir  string
+	env  []string
+	exec execer
+}
+
+// ConfigOptions configures the behavior of a [Config].
+type ConfigOptions struct {
+	// Dir specifies the directory to run Git commands in.
+	// Defaults to the current working directory if empty.
+	Dir string
+
+	// Env specifies additional environment variables
+	// to set when running Git commands.
+	Env []string
+
+	// Log used for logging messages to the user.
+	// If nil, no messages are logged.
+	Log *log.Logger
+
+	exec execer
+}
+
+// NewConfig builds a new [Config] object for accessing Git configuration.
+func NewConfig(opts ConfigOptions) *Config {
+	exec := opts.exec
+	if exec == nil {
+		exec = _realExec
+	}
+
+	if opts.Log == nil {
+		opts.Log = log.New(io.Discard)
+	}
+
+	return &Config{
+		log:  opts.Log,
+		dir:  opts.Dir,
+		env:  opts.Env,
+		exec: exec,
+	}
+}
+
+// ConfigEntry is a single key-value pair in Git configuration.
+type ConfigEntry struct {
+	Key   string
+	Value string
+}
+
+// ListRegexp lists all configuration entries that match the given pattern.
+// If pattern is empty, '.' is used to match all entries.
+func (cfg *Config) ListRegexp(ctx context.Context, pattern string) (
+	func(yield func(ConfigEntry, error) bool),
+	error,
+) {
+	if pattern == "" {
+		pattern = "."
+	}
+	return cfg.list(ctx, "--get-regexp", pattern)
+}
+
+var _newline = []byte("\n")
+
+func (cfg *Config) list(ctx context.Context, args ...string) (
+	func(yield func(ConfigEntry, error) bool),
+	error,
+) {
+	args = append([]string{"config", "--null"}, args...)
+	cmd := newGitCmd(ctx, cfg.log, args...).
+		Dir(cfg.dir).
+		AppendEnv(cfg.env...)
+
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, fmt.Errorf("stdout pipe: %w", err)
+	}
+
+	if err := cmd.Start(cfg.exec); err != nil {
+		return nil, fmt.Errorf("start git-config: %w", err)
+	}
+
+	log := cfg.log
+	return func(yield func(ConfigEntry, error) bool) {
+		// Always wait for the command to finish when this returns.
+		// Ignore the error because git-config fails if there are no matches.
+		// It's not an error for us if there are no matches.
+		defer func() {
+			_ = cmd.Wait(cfg.exec)
+		}()
+
+		// With the --null flag, output is in the form:
+		//
+		//	key1\nvalue1\0
+		//	key2\nvalue2\0
+		scan := bufio.NewScanner(stdout)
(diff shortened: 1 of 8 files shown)


abhinav:

- #326 ◀
    - #327

abhinav:
This is supposed to be user facing except there are no actual options yet, so no changelog entries to add.

internal/git/config.go
```diff
@@ -0,0 +1,145 @@
...
+		opts.Log = log.New(io.Discard)
+	}
+
+	return &Config{
+		log:  opts.Log,
+		dir:  opts.Dir,
+		env:  opts.Env,
+		exec: exec,
+	}
+}
```
ccoVeille, line 55: I would expect an Option pattern here and not this

More information here

https://dev.to/kittipat1413/understanding-the-options-pattern-in-go-390c

Your code is fine, but the option pattern will make everything simpler

internal/spice/config.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	entries, err := cfg.ListRegexp(ctx, `^`+_configNamespace+`\.`)
+	if err != nil {
+		return nil, fmt.Errorf("list configuration: %w", err)
+	}
+
+	items := make(map[string][]string)
+
+	err = nil // TODO: use a range loop after Go 1.23
```
ccoVeille, line 83: You used return on err not nil, so I don't get this


```suggestion
	// TODO: use a range loop after Go 1.23
```

internal/spice/config.go
```diff
@@ -0,0 +1,143 @@
...
+// For slice fields, all values are combined.
+//
+// If a flag is passed on the CLI, it takes precedence over the configuration.
+type Config struct {
+	// items is a map from configuration key (without the "spice." prefix)
+	// to list of values for that field.
+	items map[string][]string
+}
+
+// ConfigOptions spceifies options for the [Config].
```
ccoVeille, line 63:
```suggestion
// ConfigOptions specifies options for the [Config].
```

internal/spice/config.go
```diff
@@ -0,0 +1,143 @@
...
+		key, ok := strings.CutPrefix(entry.Key, _configNamespacePrefix)
+		if !ok {
+			// This will never happen if git config --get-regexp
+			// behaves correctly, but it's easy to handle.
+			return true
+		}
+
+		items[key] = append(items[key], entry.Value)
+		return true
+	})
```
ccoVeille, line 99: Maybe this?

```suggestion
		// ok=false will never happen if git config --get-regexp
		// behaves correctly, but it's easy to handle.
		key, ok := strings.CutPrefix(entry.Key, _configNamespacePrefix)
		if ok {
			items[key] = append(items[key], entry.Value)
		}

		return true
	})
```

internal/spice/config.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/alecthomas/kong"
+	"github.com/charmbracelet/log"
+	"go.abhg.dev/gs/internal/git"
```
ccoVeille, line 11:
```suggestion

	"go.abhg.dev/gs/internal/git"
```

internal/spice/config.go
```diff
@@ -0,0 +1,143 @@
...
+				return kong.JoinEscaped(values, flag.Tag.Sep), nil
+			}
+
+			return nil, fmt.Errorf("key %q has multiple values but no separator is defined", key)
+		}
+
+		// Last value wins if there are multiple instances
+		// for a single-valued flag.
+		return values[len(values)-1], nil
+	}
```
ccoVeille, line 142: I would have written

```suggestion
	values := c.items[key]
	switch len(values) {
	case 0:
		return nil, nil

	case 1:
		return values[0], nil
	}

	if flag.IsSlice() {
		if flag.Tag.Sep != -1 {
			// If there are multiple values, and a separator is defined,
			// let Kong split the values.
			return kong.JoinEscaped(values, flag.Tag.Sep), nil
		}

		return nil, fmt.Errorf("key %q has multiple values but no separator is defined", key)
	}

	// Last value wins if there are multiple instances
	// for a single-valued flag.
	return values[len(values)-1], nil
```

internal/git/config.go
abhinav: Thanks for the tip, @ccoVeille. I don't think the boilerplate/cost of functional options is worth it here given that the list of callers is tightly controlled—seeing as it's an internal package. The API needs to at least be public to justify the cost of functional options. (FWIW, other entry points in this package also use Options structs instead of functional options.)

internal/spice/config.go
abhinav: I've been going for two groupings: stdlib in one group, everything else in the other.
For more granular import grouping, I would want the standard tooling to support something better.
(This was actually a long-discussed thing back at a past job. We used to prefer the ordering: stdlib, project-level, external imports, but the tooling didn't do it by default so it was a constant "import grouping" comment in code reviews. We eventually decided to adjust guidance to better match tooling output.)

internal/spice/config.go
abhinav: Thanks!

internal/spice/config.go
abhinav: So the `err` is being filled in from the iteration loop below:

```
err = nil
entries(func(..., fooErr error) bool {
  if fooErr != nil {
    err = fooErr
    return false
  }
  // ...
})
```

Once Go 1.23 is released, this will all become:

```
for entry, err := range entries {
  if err != nil {
    return fmt.Errorf("read configuration: %w", err)
  }
  // ...
}
```

internal/spice/config.go
abhinav: Hm, yeah I wouldn't mind turning that into one return. Thanks!

internal/spice/config.go
abhinav: Yeah, that matches [reduce nesting](https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting), which I tend to default to in my code.

abhinav (commented):
Thanks for the review, @ccoVeille!

internal/spice/config.go
ccoVeille: Oh, that's right. Thanks for the explanation

internal/spice/config.go
ccoVeille: OK noted

ccoVeille (approved):

PR #327, opened 2024-08-07
config: Support custom shorthands

### LINKED ISSUE #324: Config: Custom shorthand support
Add support for defining custom command aliases and shorthands to augment the buitlins.

Depends on #16

abhinav:
Resolves #324

### DIFF
--- a/doc/src/cli/shorthand.md
+++ b/doc/src/cli/shorthand.md
@@ -8,9 +8,8 @@ description: >-
 
 # CLI shorthands
 
-git-spice offers short versions of most commands
+git-spice comes built-in with short versions of most commands
 to make them easier to remember and type.
-
 To determine the shorthand for a command,
 run the command with the `--help` flag.
 
@@ -21,7 +20,7 @@ Usage: gs branch (b) create (c) [<name>] [flags]
 {gray}# ...{reset}
 ```
 
-The shorthand is built by joining the bits in parentheses.
+The shorthand for a command is the bits in parentheses joined together.
 For example, the shorthand for $$gs branch create$$ above is `gs bc`.
 Another example:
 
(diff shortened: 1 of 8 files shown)


abhinav:

- #326
    - #327 ◀

ccoVeille (approved):

internal/spice/config.go
```diff
@@ -58,9 +61,13 @@ type Config struct {
...
 	// to list of values for that field.
 	items map[string][]string
+
+	// shorthands is a map from shorthand to the list of arguments
+	// that that it expands to.
+	shorthands map[string][]string
 }
 
-// ConfigOptions specifies options for the [Config].
+// ConfigOptions spceifies options for the [Config].
```
ccoVeille, line 70:
```suggestion
// ConfigOptions specifies options for the [Config].
```

doc/src/cli/shorthand.md
```diff
@@ -21,7 +20,7 @@ Usage: gs branch (b) create (c) [<name>] [flags]
 {gray}# ...{reset}
 ```
 
-The shorthand is built by joining the bits in parentheses.
+The shorthand for a command is the bits in parentheses joined together.
 For example, the shorthand for $$gs branch create$$ above is `gs bc`.
```
ccoVeille, line 24: Why using $$ ?

Isn't the LaTeX syntax operator?

I would have expected 


```suggestion
 For example, the shorthand for |gs branch create` above is `gs bc`.
```

internal/spice/config_test.go
```diff
@@ -160,6 +162,20 @@ func TestIntegrationConfig_loadFromGit(t *testing.T) {
...
 			}{Level: "hot"},
 		},
+		{
+			name: "Shorthands",
+			config: text.Dedent(`
+				[spice.shorthand]
+				can = commit amend --no-edit
+				wip = commit create -m \"wip: \\\"quoted\\\"\"
+				poorly = \"  # ignored (not a valid shorthand)
+			`),
```
ccoVeille, line 172: Is it needed to escape as you are using string literal ? (Backtick)


```suggestion
			config: text.Dedent(`
				[spice.shorthand]
				can = commit amend --no-edit
				wip = commit create -m "wip: \"quoted\""
				poorly = "  # ignored (not a valid shorthand)
			`),
```

internal/spice/config.go
abhinav: Oh, oops. Guess I undid the fix. Thanks for catching this.

doc/src/cli/shorthand.md
abhinav: `$$` is a LaTeX syntax, but it's not really supported in Markdown—unless you specifically enable it.
Since it's not part of the Markdown standard, but is obvious and familiar enough, it made sense to use it for a custom extension. I have a custom hook installed into mkdocs that will turn `$$gs foo bar$$` into a link to the CLI reference, and `$$spice.foo.bar$$` into a link to the configuration reference.

internal/spice/config_test.go
ccoVeille: Ok

PR #329, opened 2024-08-07
doc: Enumerate configurations under each command

abhinav:
For each command, enumerate configuration keys that affect it,
and link to the configuration reference.

This will be especially handy for config-only options
that don't have corresponding public flags.


abhinav:

- #329 ◀
    - #330

PR #330, opened 2024-08-07
Support setting GitHub URLs via configuration

abhinav:
Setting GITHUB_URL and GITHUB_API_URL environment variables
can be inconvenient.
Allow setting these values via git config if desired.

### DIFF
--- a/.changes/unreleased/Added-20240807-075520.yaml
+++ b/.changes/unreleased/Added-20240807-075520.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'auth: Allow setting GitHub URLs with git-config.'
+time: 2024-08-07T07:55:20.266498-07:00
--- a/doc/includes/cli-reference.md
+++ b/doc/includes/cli-reference.md
@@ -12,6 +12,8 @@ gs (git-spice) is a command line tool for stacking Git branches.
 * `-C`, `--dir=DIR`: Change to DIR before doing anything
 * `--[no-]prompt`: Whether to prompt for missing information
 
+**Configuration**: [spice.forge.github.url](/cli/config.md#spiceforgegithuburl), [spice.forge.github.apiUrl](/cli/config.md#spiceforgegithubapiurl)
+
 ## Shell
 
 ### gs shell completion
--- a/doc/src/cli/config.md
+++ b/doc/src/cli/config.md
@@ -30,6 +30,20 @@ or at the repository level with the `--local` flag.
 
 ## Available options
 
+### spice.forge.github.apiUrl
+
+URL at which the GitHub API is available.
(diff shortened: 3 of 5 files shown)


abhinav:

- #329
    - #330 ◀

messense:
It'd be really nice if `git-spice` can infer GHE server and API urls from git remote URL in `.git/config` when it's not `github.com`.

abhinav:
> It'd be really nice if `git-spice` can infer GHE server and API urls from git remote URL in `.git/config` when it's not `github.com`.

I agree, but unfortunately, unless there's a clear way to identify GHE servers (e.g. they must always have github.com in their name), that's not possible.
While git-spice currently only supports GitHub, I've intentionally tried to keep it forge-agonistic.
GitHub-specific details are in [a single place](https://github.com/abhinav/git-spice/tree/main/internal/forge/github).
Everything outside that assumes that a repository could be using any forge.
It is possible and conceivable to add support for other code forges to git-spice.
In fact, most of the CLI behavior tests run against a [fake forge with a simpler API](https://github.com/abhinav/git-spice/tree/main/internal/forge/shamhub) that is [registered only for tests](https://github.com/abhinav/git-spice/blob/23351a743c51c1234977b788486130fff4d9eca5/script_test.go#L44).

messense:
Maybe there should be an option to configure the git server type? Then you can infer the API urls based on server type and remote URL, it's just nicer to not have to explicitly write out multiple URLs.

abhinav:
> Maybe there should be an option to configure the git server type? Then you can infer the API urls based on server type and remote URL, it's just nicer to not have to explicitly write out multiple URLs.

Oh, hm. 🤔  Are you suggesting a way to say, "assume all remotes are GitHub"?
I can see that being more convenient than the version in this PR.

abhinav:
Oh, I think I understand what you mean.
The configuration would just state that this is a GHES server, and (I guess?) the API URL for those is always `/api` on the instance URL?

messense:
> The configuration would just state that this is a GHES server, and (I guess?) the API URL for those is always `/api` on the instance URL?

exactly, I don't think they are mutually exclusive, it's just a shortcut IMO.

abhinav:
Created https://github.com/abhinav/git-spice/issues/336.

Separately, @messense, re this:

>  it's just nicer to not have to explicitly write out multiple URLs.

One obvious convenience we can add here: if GitHub URL is not github.com, then API URL is GitHub URL + "/api". That would help here a little, yes?

messense:
> One obvious convenience we can add here: if GitHub URL is not github.com, then API URL is GitHub URL + "/api". That would help here a little, yes?

Yes, I think so.

PR #333, opened 2024-08-09
submit: Support navigation comment opt-out

### LINKED ISSUE #289: Config option to prevent stack info printing in PR description
Different people on my team use different stacking tools (or no tool and do it manually), so we have a github action that adds the stack information to the PR description. 

It would be nice to have a config option to prevent `git-spice` from adding the stack info, because it ends up being redundant in my case.

Depends on #16

abhinav:
Allow opting out of navigation comments in change requests,
or only posting them when there are multiple change requests.

This option may be set with the `--nav-comment` flag,
or the `spice.submit.navigationComment` configuration key.

Resolves #289

### DIFF
--- a/internal/git/config.go
+++ b/internal/git/config.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/charmbracelet/log"
 )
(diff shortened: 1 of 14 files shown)


abhinav:

- #333 ◀

internal/git/config.go
```diff
@@ -54,9 +55,79 @@ func NewConfig(opts ConfigOptions) *Config {
 	}
 }
 
+// ConfigKey is divided into three parts:
+//
+//	section.subsection.name
+//
+// subsection may be absent, or may be comprised of multiple parts.
+// section and name are case-insensitive.
+// subsection is case-sensitive.
```
abhinav, line 64: This was fun to discover

Issue #336, opened 2024-08-09
Configurable forge kinds


abhinav:
Right now, the way we determine the code forge is by matching the remote URL.

For GitHub, it must be github.com, or whatever is configured with GITHUB_URL.

Per discussion in #330, a better UX for GHES would be the ability to specify that they're using a specific kind of forge (e.g. GHES), and then the GitHub URL is determined based on the Git remote, and the API URL guessed based on that.

PR #337, opened 2024-08-09
forge/github: Determine API URL from base URL if not set

abhinav:
For GHES, usually the API URL is /api under the base URL.
So if the API URL is not set, and base URL is not github.com,
we can assume the API URL is $baseURL/api.

Users can still override this with the `spice.forge.github.apiUrl`
or `GITHUB_API_URL` environment variable.

Follow up to https://github.com/abhinav/git-spice/pull/330#issuecomment-2277077683
Relates to #336


abhinav:

- #337 ◀

PR #349, opened 2024-08-16
onto: Apply even if branch is already on target

### LINKED ISSUE #321: {branch, upstack} onto: restack even if no-op
There may be an expectation from users that `branch onto foo` will always move the current branch and its commits on top of foo.
Right now, if the current branch is already stacked on top of foo, but has diverged (and needs restacking), we'll no-op.

It is preferable that `branch onto` rebases the current commits on top of foo even foo is already the current branch's base.
The same may be desirable for `upstack onto`.

abhinav:
The onto operations do two things:

- update internal state of the branch to target a new base
- rebase commits $oldBase..$branch onto $newBase

There's an expectation from users that if branch is already on target,
the command should still rebase the commits if the target is ahead of
the branch.

This change makes the onto operations always run,
even if the branch is already on target.

Resolves #321

### DIFF
--- a/.changes/unreleased/Changed-20240816-054428.yaml
+++ b/.changes/unreleased/Changed-20240816-054428.yaml
@@ -1,0 +1,5 @@
+kind: Changed
+body: >-
+  {branch, upstack} onto: Rebase commits even if the branch's base matches the target.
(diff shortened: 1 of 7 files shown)


abhinav:

- #349 ◀

PR #354, opened 2024-08-22
submit: --no-publish should work for unsupported forges

abhinav:
If using `submit` commands with --no-publish,
there's no need to require remote repository information.
As long as we can `git push` to it, we're good.

This commit reorganizes the information retrieval in submit
commands to only fetch the remote repository information
when necessary.

Resolves #351

PR #355, opened 2024-08-22
submit: support --publish opt-in mode

abhinav:
Change --no-publish to --publish, defaulting to true,
with a --no-publish opt-out.
Support setting the default value via configuration,
allowing users to always opt out of creating CRs,
unless the --publish flag is added to the command.

This will make the experience of using submit with unsupported forges
smoother.


abhinav:

- #354
    - #355 ◀

PR #363, opened 2024-08-28
git: Don't use rev-list's --no-commit-header

abhinav:
This flag was introduced in Git 2.33,
which is newer than the default on many systems.
We don't absolutely need this flag -- just manually skip the header.


abhinav:

- #363 ◀

Issue #369, opened 2024-08-30
Cached PR template does not update after remote file changed


messense:
I've updated `pull_request_template.md` in trunk of remote git repository, but `git-spice` still uses the old cached template when submitting stacks.

abhinav:
Hey, just to narrow this down: Is the branch you're submitting from restacked on top of trunk?

messense:
Yes.

abhinav:
Follow up: Is trunk fully synced?
If not, that would explain why it's using the old cache key because right now, it checks `$trunk`, not `origin/$trunk`, so even if an update to the template is merged to trunk, and you've fetched trunk, the local ref `$trunk` is behind and the cache is seeing the wrong version.
Easily fixable, but I want to make sure that's what you're seeing.

abhinav:
If the above is the issue, then #370 should fix this.

messense:
> Is trunk fully synced?

I think so, I've run `git pull` on my local checkout of trunk branch and also run `gs repo sync` again.

abhinav:
Oh, `origin/$trunk` was in sync? That doesn't sound right.
Would you be able to try with #370? It's merged on main, but you'd have to [build from source](https://abhinav.github.io/git-spice/start/install/#build-from-source).
(And if the issue still persists: what is the path to the template file?)

messense:
> The other thing I'm guessing could be wrong here is that the template file name is lower case, and right now, git-spice only checks upper-case names (PULL_REQUEST_TEMPLATE).

yes, the file name is indeed lower case: `.github/pull_request_template.md`

abhinav:
Aha! That's it! Fix incoming!

abhinav:
Alright, https://github.com/abhinav/git-spice/pull/371 should take care of it.

messense:
Thanks for the quick fix! v0.5.2 works fine now.

PR #370, opened 2024-08-30
submit: Invalidate template cache if origin/main is updated

abhinav:
We calculate the template cache key based on the hashes of the files
inside the locations in the repository where tempates are stored,
but we check their state at `$trunk` instead of `$remote/$trunk`.

This results in the cache not being invalidated
if the remote has been updated and fetched, but the local ref is behind.

Ref #369


abhinav:

- #370 ◀

PR #371, opened 2024-08-30
submit: Fix outdated PR templates with lower-case names

### LINKED ISSUE #369: Cached PR template does not update after remote file changed

abhinav:
We use the contents of possible PR template sites
as the hash key for caching the PR templates we get from the remote.
This was only looking at `PULL_REQUEST_TEMPLATE.md` names,
even though `pull_request_template.md` is valid and accepted.

This change ensures that lower-cased versions are also considered
when calculating the hash key.

To do this, we consider upper-case and lower-case variants
of configured Forge template paths when calculating the cache key.
Case-insensitivity of this path is considered part of the Forge contract
because GitHub and GitLab both treat these as case-insensitive-ish.
We can adjust this assumption if we add a forge that doesn't.

Resolves #369


abhinav:

- #371 ◀

abhinav:
Hm, actually I don't like this version.
GitLab Merge Request templates are case-insensitive.
I'll encode this in the contract.

PR #387, opened 2024-09-07
Windows support

abhinav:
This adds support for using git-spice with Windows.
I'm pretty sure it already works, but testing it is a bit difficult,
especially since I don't have a Windows system to test on.

All terminal prompt use creack/pty to create a fake terminal.
Until creack/pty#155 is merged, these tests cannot be run on Windows.
We'll skip the prompt tests on Windows for the time being.


abhinav:

- #387 ◀

PR #401, opened 2024-09-14
repo sync: Fix deleting wrong tracking branch for renamed branches

abhinav:
If a branch is renamed after submission and then merged,
the tracking branch name does not match the local branch name.

This case must be correctly handled when deleting the tracking branch:
if we know the upstream branch name use `$remote/$upstreamBranch`
instead of `$remote/$branch`.

Also drop the outdated comment about how we detect merged branches
sinc we're indeed doing both (1) and (2) now.


abhinav:

- #401 ◀

PR #402, opened 2024-09-14
ci: Test with different version of Git

abhinav:
If we're going to add support for things like git-replay,
we'll end up needing to test against different versions of Git.
Set this up so that we can also establish a minimum required version.

Tests had to change how they injected custom Git configuration.
Support for `GIT_CONFIG_KEY_<N>` was added in 2.31,
so we're just going to fill a `GIT_CONFIG_SYSTEM` with values instead.


abhinav:

- #402 ◀

Issue #391, opened 2024-09-10
'gs repo sync' error for on-premise GitLab instances


gr4cza:
`gs repo sync` returns 
```
INF main: already up-to-date
ERR Could not guess repository from remote URL url=https://gitlab.xyz.io/abc.git
ERR Are you sure the remote identifies a supported Git host?
FTL gs: unsupported Git remote URL
``` 
for on-premise GitLab instances. It would be really nice, if it could handle custom urls without errors.

abhinav:
Hello! Thanks for reporting. This is worth fixing.

Note that without the ability to check for merged CRs (without #348), `gs repo sync` will probably just do `git pull`, but it's a reasonable graceful degradation.
There may be some room for local merge detection, but I suspect it won't handle cases like squash-merged CRs.

Just out of curiosity: Are you using `submit` with `--no-publish` (or the [spice.submit.publish](https://abhinav.github.io/git-spice/cli/config/#spicesubmitpublish) configuration) to manage the stack?

gr4cza:
Hi! First of all, thank you for your amazing work!

Yes, i'm using it with the `spice.submit.publish` flag set to `false`. It is really handy to force push multiple branches at the same time.

PR #404, opened 2024-09-15
repo sync: Gracefully handle unsupported forges

abhinav:
When `repo sync` is run on a repository with a remote
that is not a known forge (e.g. GitLab),
instead of failing with an error because we can't detect merged PRs,
just log a warning and skip the branch deletion step.

This is only a partial fix for #391.
It'll still leave the merged branches around,
and worse, the upstack branches will still point to merged branches
as their bases.
Since we can easily detect merged branches that used a merge commit,
we will do that instead of a warning in a follow-up PR.


abhinav:

- #404 ◀

PR #405, opened 2024-09-15
refac(repo sync): split out branch deletion logic

abhinav:
Separate searching for branches to delete from actually deleting them.
This will allow re-using the same logic to delete branches
that were merged for unsupported forges.
This is also necessary to later make the deletion logic smarter
to delete branches in the correct order.


abhinav:

- #405 ◀
    - #406

PR #415, opened 2024-09-22
chore: update GPL license ref

chenrui333:
it should be GPL-3.0-or-later rather than GPL-3.0


abhinav (approved):
Oops, yeah. Thanks for fixing!

PR #416, opened 2024-09-22
ci: update to use `go-version-file` rather than `go-version`

chenrui333:
using go-version-file would ease the go version management, which takes go.mod as source of truth.


abhinav (approved):
Thanks!

PR #422, opened 2024-09-26
state: Lazy branch graph load, incremental changes

abhinav:
Instead of building and discarding the branch graph for UpdateBranch,
provide a BranchTx API to prepare changes to the branch state.

Required state will be loaded lazily as needed for the underlying checks.
Changes will be accumulated in the BranchTx until Commit is called.

For now, the old UpdateBranch method just wraps BranchTx.
We'll delete it in the future and use BranchTx directly everywhere.

[skip changelog]: No user facing changes.J ust a better version of #419.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,6 @@
 *.txt text eol=lf
 
 *.png filter=lfs diff=lfs merge=lfs -text
+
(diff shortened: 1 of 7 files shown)


abhinav:

- #422 ◀

PR #423, opened 2024-09-27
branch create: Don't create or commit if store would reject

abhinav:
By using the new BranchTx API,
we're able to check whether the store will accept the change
before we actually create the branch.

So, we now first prepare the transaction with the new branch state,
and only if that succeeds do we actually create the branch.

This way, if the store would reject the change
because the base branch is untracked or something similar,
we'll roll back to the original target branch and staged changes.


abhinav:

- #423 ◀

PR #432, opened 2024-09-30
branch selector auto-focus based on fuzzy ranking

### LINKED ISSUE #430: request: branch selector auto-focuses on the most relevant branch when searching
Hi! I have been daily driving git-spice for a while and noticed that when I do `gs branch checkout` or anything that opens the branch selector widget, and then I try to search for a branch name, the cursor does not focus on the most relevant branch matching my search term -- it focuses on the very top branch or the current branch in the selector widget instead, which is fine for the most part, but I expect for, say, the most relevant one to appear first (e.g matched by prefix)

For instance, I have this stack tree, and I'm currently on branch `marketing`:
```
main <- marketing (current) <- marketing-2
```

If I search for "main", it would be like this (the [..] are the characters that get highlighted):

```
[main] <- [ma]rket[in]g (current) <- [ma]rket[in]g-2
```

But my cursor is still on `marketing`, meanwhile the `main` branch matched my search term completely; I would expect in that case the `main` branch should be focused i.e:

```
[main] (current) <- [ma]rket[in]g <- [ma]rket[in]g-2
```

It also doesn't have to be an exact match to be focused; something like if I search for "mai", then I should expect `main` to be focused because "mai" is a prefix of `main`, as opposed to the other two branches `marketing` and `marketing-2`.

Let me know if this is something we can implement; I'm open to work on this.

snokpok:
resolves #430; I used `fuzzy` to rank the selectable branches


abhinav (approved):
Seems to work great. Thanks!

PR #448, opened 2024-10-25
fix: correct navigation comment selection

### LINKED ISSUE #447: github: submit comment detection picks up the wrong comment
When you call `gs branch submit` on a PR not previously submitted by git-spice,
it will attempt to detect existing gs navigation comments by looking for these patterns:

https://github.com/abhinav/git-spice/blob/38756dfcfc03de56e7c75283c7240cd755bb50c6/submit.go#L363-L365

There's a bug in the underlying comment filtering logic:

https://github.com/abhinav/git-spice/blob/38756dfcfc03de56e7c75283c7240cd755bb50c6/internal/forge/github/comment.go#L207-L212

The `continue` is supposed to skip the outer loop (the comment), not the inner (the filters).
As a result of this, the logic currently always picks the first comment in the PR as the navigation comment.

yomapeed:
This PR fixes a bug in the loop where we always select the first comment and don't filter anything due to a misplaced `continue`.

fixes #447

### DIFF
--- a/internal/forge/github/comment.go
+++ b/internal/forge/github/comment.go
@@ -205,10 +205,15 @@ func (f *Repository) ListChangeComments(
 			}
 
 			for _, node := range q.Node.PullRequest.Comments.Nodes {
+				isNavigationComment := true
 				for _, filter := range filters {
 					if !filter(node) {
-						continue
+						isNavigationComment = false
+						break
 					}
+				}
+				if !isNavigationComment {
+					continue
 				}
 
 				item := &forge.ListChangeCommentItem{
--- a/internal/forge/github/comment.go
+++ b/internal/forge/github/comment.go
@@ -205,14 +205,14 @@ func (f *Repository) ListChangeComments(
 			}
 
 			for _, node := range q.Node.PullRequest.Comments.Nodes {
-				isNavigationComment := true
+				match := true
 				for _, filter := range filters {
 					if !filter(node) {
-						isNavigationComment = false
+						match = false
 						break
 					}
 				}
-				if !isNavigationComment {
+				if !match {
 					continue
 				}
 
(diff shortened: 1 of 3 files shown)


abhinav (approved):
Thanks!
Would you mind making the following changes?

- Please add a new changelog entry with `changie new`. (You may need to install it. I'll look into making that easier in the repo setup in the future.)
- If you find an easy way to do so, please also add a regression test. It's probably possible with a fake GraphQL client or HTTP client that returns a static string as the JSON response.

No pressure to make these changes if you don't have the time.
I'll probably be able to make time to amend this PR with those changes this weekend and merge this PR.

internal/forge/github/comment.go
```diff
@@ -205,11 +205,16 @@ func (f *Repository) ListChangeComments(
 			}
 
 			for _, node := range q.Node.PullRequest.Comments.Nodes {
+				isNavigationComment := true
```
abhinav, line 208: one nit: this isn't necessarily a navigation comment. At this layer of abstraction, it's just a "match".

```suggestion
				match := true
```

internal/forge/github/comment.go
```diff
@@ -205,11 +205,16 @@ func (f *Repository) ListChangeComments(
...
 				for _, filter := range filters {
 					if !filter(node) {
-						continue
+						isNavigationComment = false
```
abhinav, line 211:
```suggestion
						match = false
```

internal/forge/github/comment.go
```diff
@@ -205,11 +205,16 @@ func (f *Repository) ListChangeComments(
...
+						break
 					}
 				}
+				if !isNavigationComment {
```
abhinav, line 215:
```suggestion
				if !match {
```

abhinav:
Okay, it looks like the github package doesn't have an existing test that uses hand-written responses in the way I'm suggesting. There are [tests in graphqlutil](https://github.com/abhinav/git-spice/blob/main/internal/graphqlutil/error_test.go) that do that, but given that there isn't an easy set up for testing that in this package already, I'm happy to add the regression test myself before merging this PR.

yomapeed:
> * Please add a new changelog entry with `changie new`. (You may need to install it. I'll look into making that easier in the repo setup in the future.)


Change log has been added.

PR #455, opened 2024-11-02
fix(branch submit): handle base branch renaming

abhinav:
When submitting a branch, we already handle the case
where the local branch has been renamed since submitting.
The original upstream name is still used.

But we did not account for that change
when submitting branches based on top of the renamed branch.

This fixes that--just a matter of plumbing the right variables around.


abhinav:

- #455 ◀

PR #461, opened 2024-11-08
submit: Generate unique remote name if branch name is taken

### LINKED ISSUE #408: submit: Handle remote ref name already taken
Today, all submit commands push to `$remote/$name` where `$name` is the same as the local branch name—except if the branch was renamed after the initial push.

This is fine as the default, but it doesn't handle the case where `$name` is already taken by the remote.
This will especially be true if the local name happens to be something generic.
We can probably inspect local tracking branches, and use a different name (e.g. `$name-$n` for an incrementing value of `$n`) if that's the case.

abhinav:
When submitting a branch `$name`,
it's possible that someone else has already taken that name.
For example, the local branch name might be something generic
like 'feature', and someone else may already have taken that name.

To handle this case, gs will now query the remote with `git ls-remote`,
and generate a unique name for the branch if the original name is taken.
New names start at `$name-2`, and keep incrementing the number
until a free name is found.

Making this change results in a couple other changes:

- If the branch has an upstream associated with it,
  e.g. with `git push -u origin foo:bar`, we'll prefer that name
  since the user has already pushed the branch and chosen it.
  This makes sense: I pushed the branch, and associated the upstream
  with the branch, so that's what should be used.
- When matching externally submitted CRs to the branch,
  just matching on the branch name is no longer sufficient.
  If we're handling conflicting names, we have to account for the case
  where someone else submitted a CR with the conflicting branch name.
  Therefore, CR matching now also requires that the HEADs match.

Resolves #408

### DIFF
--- a/.changes/unreleased/Fixed-20241106-052308.yaml
+++ b/.changes/unreleased/Fixed-20241106-052308.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'submit: If a branch has already been pushed to the remote repository with `git push -u`, use that branch name when creating a CR.'
+time: 2024-11-06T05:23:08.585547-08:00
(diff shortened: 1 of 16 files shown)


abhinav:

- #461 ◀

PR #463, opened 2024-11-09
submit: list templates: allow configuring timeout

abhinav:
The hard-coded timeout for looking up CR templates
was causing issues for users with slow connections.

```
WRN Could not list change templates error="list templates: Post \"https://api.github.com/graphql\": context deadline exceeded"
```

This adds a `spice.submit.listTemplatesTimeout` configuration
to configure this timeout.
Setting the timeout to 0 disables it completely.

Resolves #438

Issue #438, opened 2024-10-04
Template Cache Warnings


klnusbaum:
I'm getting a variety of Warnings related to the "templates". Some examples are:

```
WRN Failed to cache templates err="cache templates: update tree: make root tree: start: context deadline exceeded"
```

```
WRN Failed to cache templates err="cache templates: write object: hash-object: signal: killed"
```

```
WRN Could not list change templates error="list templates: Post \"https://api.github.com/graphql\": context deadline exceeded"
```

I'm not entirely sure what's going on. For further context:

```
# git cat-file -p refs/spice/data:templates
{
  "key": "ef63c6f2c35943a1d97e383915d2f4a88fbef85afaeae105aa8bbfcd97fb41bc",
  "templates": []
}
```

abhinav:
Confirming a couple things:

- Does your repository have PR templates? How big are they?
- The "Could not list" and "Failed to cache" warnings: did they appear together or in separate instances?

Context: We look up the templates in the background with a timeout of just a second:
https://github.com/abhinav/git-spice/blob/e10a9029bdfbac302e154e075665f5cd37d31078/branch_submit.go#L590-L593

That would cause all of the warnings above to occur.

klnusbaum:
These two errors appeared together just now when I was submitting a stack
```
2024-10-07 09:29:06 PDT infrastructure@kdevenv # gs ss -c
WRN Failed to cache templates err="cache templates: update tree: make root tree: start: context deadline ex
ceeded"
INF Created #7667: ... [redacted]
WRN Failed to cache templates err="cache templates: write object: hash-object: signal: killed"
INF Created #7668: ... [redacted]
```

klnusbaum:
Based on [these instructions](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) I don't think we have any PR templates, but I could be wrong.

abhinav:
Someone else also reported this. I think we can bump the timeout and/or make it configurable with, say `spice.submit.templateLookupTimeout` or something like that.

PR #464, opened 2024-11-09
submit: Use cached templates if forge times out

abhinav:
Follow up to #463

If the template lookup with the forge times out,
but we have cached templates sitting around,
there's no reason to present "no templates" to the user.

Use the cached templates, just let them know they're outdated.

Refs #438


abhinav:

- #464 ◀

Issue #469, opened 2024-11-11
`gs` conflicts with existing command (ghostscript)


guyboltonking:
When installing using homebrew, I get a conflict because I have ghostscript (which has been known as `gs` for a very long time).  Would it be possible to choose a different name?  I've worked around this by symlinking the binary as git-spice into my `~/bin` (and changing the completions using:

```
git-spice shell completion fish | sed -e 's/__complete_gs/__complete_git_spice/' -e 's/-c gs/-c git-spice/' | source
```

...), but I imagine it will hit other people.

dvogt23:
yes, for me as well. Would be great to be able to create a git subcommand for spice as [git absorb](https://github.com/tummychow/git-absorb) doing it.

abhinav:
Hey folks, thanks for creating the issue.

I understand that the ghostscript conflict is inconvenient. I'm willing to amend the installation instructions with a warning and a workaround for ghostscript users, but I have a strong preference for keeping the current name. Renaming is not an option I'm currently open to. The current name and direct invocation (as opposed to a git subcommand) better matches the user experience I want from this tool.

RE: fixing completions with `sed`:
Oh, that shouldn't be necessary. The command should use its own name even when renamed. Let me fix that.

abhinav:
> In terms of a workaround, could you lean on [homebrew options](https://docs.brew.sh/Formula-Cookbook#adding-optional-steps) to provide a `--with-git-subcommand` option (so the default is your preferred UX, but the workaround works _within_ homebrew instead of needing a manual symlink step)?

I considered that, and I don't know how strict the policy is, but per that section of the Formula Cookbook:

> Note: options are not allowed in Homebrew/homebrew-core as they are not tested by CI.

Feel free to try; if they accept the PR, I have no opposition to an install-time flag for this.

abhinav:
Oh, I thought you were talking about the existing Homebrew formula.
git-spice was added to homebrew-core in September (https://github.com/Homebrew/homebrew-core/pull/191536).
I hadn't updated the installation instructions to refer to that because I hadn't gotten the automated `brew bump-formula-pr` workflow working right until, well, [today](https://github.com/abhinav/git-spice/actions/runs/11798296278/job/32864851024).

For my Homebrew tap: [goreleaser](https://goreleaser.com/customization/homebrew/) generates and pushes that formula. Goreleaser doesn't seem to have a means of adding options, _but_, it supports publishing multiple formulas.

So, I guess we have two options:

[...]

PR #470, opened 2024-11-12
completion: recognize renamed binary

abhinav:
If the command is renamed, don't use "gs" as the name in help
or in the completion script. Use the renamed command name instead.

e.g.

```
❯ make bin/gs
❯ mv bin/gs gsp
❯ ./gsp shell completion bash
complete -C /.../src/git-spice/gsp gsp
```

Refs #469


abhinav:

- #470 ◀

PR #480, opened 2024-11-24
ui: New terminal prompting API

abhinav:
We currently emulate a full terminal with mideterm in test scripts.
This gets a bit finicky when things break because test scripts
have limited control flow.
Those tests also don't run on Windows because the pty library
does not yet have support for Windows consoles.
Furthermore, the terminal tests aren't used at all for unit testing
the terminal widget implementations because they're too high-level.

This change attempts to start addressing these issues by refactoring
how we render to the terminal and test it.

To start with, a new `View` interface is introduced
that provides write-only access to the terminal.
It's just an `io.Writer` at this level.

The key is the new `InteractiveView` interface
that provides a request-response style method
for prompting the user for input with a few `ui.Field`s.
A given `View` can be upgraded to `InteractiveView`
if the application is currently running in interactive mode.

`InteractiveView` is implemented by the `TerminalView` struct which runs
the current normal rendering logic.

`InteractiveView` is also implemented by a new `uitest.RobotView`
which answers prompts with values from a fixture file
and renders the widgets back into the fixture file for testing.
So the idea is that in script tests, when we expect to be prompted,
we'll provide a fixture file that contains inputs to those prompts.
The `RobotView` prints a copy of the input fixture file
and a rendering of each prompt to an output file.
The script test will be able to compare the output file
with the original input file to verify that the prompts were rendered
correctly.

This new design has a few nice properties:

- It will likely speed up script tests because we are doing less.
  There won't be any need to sleep until something is rendered,
  for example.
- It will make script tests that need to prompt much easier to write.
- Code that writes to the terminal will now be more explicit
  because it must now have a `View` argument instead of just using
  `os.Stderr` will-nilly.
- Script tests will drop the dependency on creack/pty so all prompt
  tests will be able to run on Windows too.

While the script tests won't use midterm terminal emulation anymore,
that will still be used for unit testing terminal widgets directly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,10 +76,14 @@ make cover
 
 ### Test scripts
 
-Tests for the project make heavy use of the
-[testscript package](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript).
+Tests for the project make heavy use of the go-internal/testscript package.
+Read more about the test script language at [testscript package](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript).
+
 Tests scripts are stored inside the testdata/script directory.
-Read more about the test script format in the documentation of the package.
+On top of the base functionality of testscript,
+we add a bunch of custom commands and infrastructure to make testing easier.
+
+Read more about these in the [test scripts README](testdata/script/README.md).
 
 ## Releasing a new version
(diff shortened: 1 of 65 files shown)


abhinav:

- #480 ◀

PR #481, opened 2024-11-24
test scripts: use robot view for testing prompts

abhinav:
Switch a few of the test scripts that use prompts
to the new RobotView introduced in #480.

These tests can now run on Windows.


abhinav:

- #480
    - #481 ◀

Issue #485, opened 2024-11-25
Release GitLab support


abhinav:
@gr4cza added GitLab support to git-spice in #477.
The following items remain before releasing it:

- [x] See if the OAuth2 scopes can be narrowed down. #489
- [x] Update documentation website #487

abhinav:
Docs are ready to merge. Closing this.

PR #477, opened 2024-11-19
feat: Add GitLab forge Integration to Git-Spice

gr4cza:
This pull request introduces support for GitLab integration, including authentication, metadata handling, comment management, change editing, and finding changes. 

For authentication, it only implements _Personal Access Token_ auth, because I wasn't able to make the [Device authorization grant flow](https://docs.gitlab.com/ee/api/oauth2.html#device-authorization-grant-flow) work on [gitlab.com](gitlab.com).

For the GitLab API integration, it uses [xanzy/go-gitlab](https://github.com/xanzy/go-gitlab) SDK.

There is a lot of duplicated code in the implementation since there are many similarities between the two forges, but I did not want to mess with the already existing code at first.

What missing:
- [ ]  Doc update
- [x]  Test copy + exstension

Feel free to modify this as needed to better fit the specifics of your work.

Resolves #348

PR #487, opened 2024-11-25
doc: Update for GitLab support

abhinav:
Update documentation website to reflect the new GitLab support
from #477.

refs #485

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 git-spice is a tool for stacking Git branches.
 It lets you manage and navigate stacks of branches,
 conveniently modify and rebase them,
-and create GitHub Pull Requests from them.
+and create GitHub Pull Requests or GitLab Merge Requests from them.
 
 See <https://abhinav.github.io/git-spice/> for more details.
 
@@ -24,7 +24,7 @@ Usage looks roughly like this:
 # Stack a branch on top of the current branch.
 $ gs branch create feat1
 
-# Stack aonther branch on top of feat1.
+# Stack another branch on top of feat1.
 $ gs branch create feat2
 
 # Submit pull requests for feat1 and feat2.
@@ -51,8 +51,8 @@ $ gs sr        # stack restack
 ## Features
 
 - Create, edit, and navigate stacks of branches with ease.
-- Create and update GitHub Pull Requests for the entire stack
-  or parts of it with a single command.
+- Submit the entire stack or parts of it with a single command.
+  Supports GitHub and GitLab.
 - Keep using your existing workflow and adopt git-spice incrementally.
 - Completely offline operation with no external dependencies
   until you push or pull from a remote repository.
--- a/doc/hooks/badge.py
+++ b/doc/hooks/badge.py
@@ -1,0 +1,44 @@
+"""
+Adds a badgge shortcode in the form:
+
+    <!-- gs:badge ICON TEXT -->
+
+Where ICON is a mkdocs-material icon name
+and TEXT is the text to display.
+"""
+
+import re
+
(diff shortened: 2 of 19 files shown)


abhinav:

- #487 ◀

abhinav:
@gr4cza Took a pass over the docs to add GitLab. 
Do you know the answer to these questions?

- Is GitLab also subject to the "squash-merges require restacking" limitation
- What are the steps for GitLab Enterprise (is that what it's called?)

(I'm happy to update the docs if I have the answers, but you can also PR against this branch if you want to add more than that.)

gr4cza:
@abhinav I just did the testing for `squash-merges require restacking` and saw that you already verified it. Anyway, I came to the same conclusion. 

For the _Enterprise_ part:
I guess the service which is interesting to us called GitLab Self-hosted. That is, when the Base URL (and the API url) differs from the main gitlab.com. There will be a problem with the OAuth authentication. In a Self-hosted GitLab there will not be any `git-spice` OAuth client set up. So first the server admin has to create a new Application. But the problem is that this application will have a different clientId from what is hardcoded to git-spice currently. So we have to make it configurable.

@abhinav I just finished testing `squash-merges require restacking` and noticed that you’ve already verified it. Anyway, I came to the same conclusion.

For the _Enterprise_ part:
I believe the service we’re interested in is called GitLab Self-hosted. That is, when the Base URL (and the API URL) differs from the main gitlab.com. In such cases, there will be an issue with OAuth authentication. In a Self-hosted GitLab instance, there won’t be any preconfigured `git-spice` OAuth client. Therefore, the server admin will first need to create a new Application. However, this application will have a different `clientId` than the one currently hardcoded into git-spice.  Because of that, I think we’ll need to make it configurable.

gr4cza:
@abhinav I just finished testing `squash-merges require restacking` and noticed that you’ve already verified it. Anyway, I came to the same conclusion.

For the _Enterprise_ part:
I believe the service we’re interested in is called GitLab Self-hosted. That is, when the Base URL (and the API URL) differs from the main gitlab.com. In such cases, there will be an issue with OAuth authentication. In a Self-hosted GitLab instance, there won’t be any preconfigured `git-spice` OAuth client. Therefore, the server admin will first need to create a new Application. However, this application will have a different `clientId` than the one currently hardcoded into git-spice.  Because of that, I think we’ll need to make it configurable.

abhinav:
> Therefore, the server admin will first need to create a new Application. However, this application will have a different clientId than the one currently hardcoded into git-spice. Because of that, I think we’ll need to make it configurable.

Ack. We can do that. It'll just be another hidden flag like this with a config key like `forge.gitlab.oauth.clientID`. 

https://github.com/abhinav/git-spice/blob/ccf453784da40f13c0ffc5174ebc960261f60a4a/internal/forge/gitlab/forge.go#L29

gitlab.Forge would get a helper `OAuthClientID()` method that returns the configured value from options or the hard-coded one, and the DeviceFlowAuthenticator would be called with that.

If you have the time to open a PR for that, that would be cool.
Otherwise, I'll have time later tonight/tomorrow. 

After that I'll update this PR with relevant documentation for GitLab Self-Hosted, and we should be good to go.

gr4cza:
I will try to do it tonight! Btw I did some research about how can we add gitlab-cli auth support as well and I think I find a way, so I will try to add that too.

abhinav:
@gr4cza I'm going to merge the doc updates. It covers Self-Hosted, OAuth in self-hosted, and other things we discuss. I think I got everything. Anything else, I'm happy to iterate on in the future.

abhinav:
image

went wild with pressing the wrong buttons there 😅

PR #488, opened 2024-11-25
fix(gitlab): don't lose MR number for listed comments

abhinav:
We list comments on a MR to find and associate navigation comments.
These need the MR number known when sending update requests.
Without it, the MR number is 0 for all comments,
so for MRs submitted with git-spice from a different machine,
we won't recover the navigation comments.

[skip changelog]: not released yet, part of the same feature


abhinav:

- #488 ◀

PR #499, opened 2024-11-27
submit: warn if a branch's base is unsubmitted

### LINKED ISSUE #460: branch submit: better error message if base hasn't been submitted
When base branch is non-trunk, but hasn't been submitted, GitHub will reject the submission in `gs branch submit` because it doesn't know the base ref and won't see any commits between base and head.

The error takes the form:

```
Head sha can't be blank, Base sha can't be blank, No commits between [...] and [...], Base ref must be a branch
```

gs branch submit should confirm that non-trunk base has been submitted before it submits the current branch, while still allowing for `--no-publish` for non-GitHub workflows.

abhinav:
Resolves #460


abhinav:

- #499 ◀

abhinav:
This isn't right. Will take a different stab later.

PR #500, opened 2024-11-27
commit create: Add --fixup flag

### LINKED ISSUE #492: Add --fixup to commit create
I use the fixup command quite a bit during my workflow. This is pretty useful during `gs be` commands, but I don't get the benefit of auto-restacking when I use `git commit --fixup abc1234`.

Is there any interest in expanding the options to commit create here to accept the fixup option?

frank-west-iii:
This adds a new --fixup flag for allowing you to create fixup commits.

--fixup takes a commitish which targets where the fixup will be applied.
--fixup also takes the search :/<text> to target the subject of a commit
message to target the correct commit.

Using gs branch edit with the autosquash option set in git allows us to
automatically fixup our commits.

Resolves #492

### DIFF
--- a/.changes/unreleased/Added-20241127-125953.yaml
+++ b/.changes/unreleased/Added-20241127-125953.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'commit create: Added a --fixup flag to allow adding a fixup commit.'
+time: 2024-11-27T12:59:53.626214-08:00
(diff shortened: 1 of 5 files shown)


frank-west-iii:
I walked away and just had a thought that this probably shouldn't be `-f` since that's not a legit flag for commit and should be the entire `--fixup` command instead. Let me take another go at that real fast.

Issue #380, opened 2024-09-04
Add `--no-verify` to commit, push, etc


ElPicador:
I'm using a lot of pre-commit/push hooks. And sometimes skipping them with a `--no-verify` can be useful.

I'm willing to implement this feature. Is it something you would be open to accept as a contribution? Any tips/comments on how to implement it?

abhinav:
Hello! Yes, a contribution for this would be welcome.

In terms of implementation:
The flag will be added to [branchCreateCmd](https://github.com/abhinav/git-spice/blob/8cc330f5fa24bfbc80fb61d8fab0c03716927eb9/branch_create.go#L14) and the three commit subcommands [amend](https://github.com/abhinav/git-spice/blob/8cc330f5fa24bfbc80fb61d8fab0c03716927eb9/commit_amend.go#L13), [create](https://github.com/abhinav/git-spice/blob/8cc330f5fa24bfbc80fb61d8fab0c03716927eb9/commit_create.go#L13), and [split](https://github.com/abhinav/git-spice/blob/8cc330f5fa24bfbc80fb61d8fab0c03716927eb9/commit_split.go#L13). All of these commands call `git.Repository.Commit`, which accepts a [git.CommitRequest](https://github.com/abhinav/git-spice/blob/8cc330f5fa24bfbc80fb61d8fab0c03716927eb9/internal/git/commit.go#L101). We'll need a new field on that struct for this option, and in [Repository.Commit](https://github.com/abhinav/git-spice/blob/8cc330f5fa24bfbc80fb61d8fab0c03716927eb9/internal/git/commit.go#L127) we can convert it into a flag passed to Git.

Other minor notes about contributing:

- The project uses [changie](https://changie.dev/) to track unreleased user-facing changes.
- After adding flags, run `make generate` inside the doc/ directory to update the CLI reference.
- Integration tests use [testscript](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript). See [testdata/script/README](https://github.com/abhinav/git-spice/blob/main/testdata/script/README.md) for details. You should be able to adapt any of the existing tests to verify this behavior end-to-end.

frank-west-iii:
@ElPicador, just was looking to see if this functionality exists. Want to check on whether or not you're actively working on this PR or not. Cheers.

abhinav:
@frank-west-iii Thanks! That's fair, the mismatch with `git commit` isn't completely intentional.
[...]

PR #502, opened 2024-11-29
commit amend: deprecate -n in favor of --no-edit

abhinav:
`git commit --no-edit` does not have a `-n` short form
so we shouldn't either.

In the interest of not breaking anyone, just deprecate the flag first.
We'll warn users when they use the short form for now.

Refs https://github.com/abhinav/git-spice/issues/380#issuecomment-2508558831


abhinav:

- #502 ◀

Issue #331, opened 2024-08-07
submit: Navigation comment: Include merged PRs


abhinav:
Navigation comment posted by `submit` commands drops a merged branches from upstack branch comments when they're updated.
That is, given `main -> A -> B -> C`, when A merges, the comments for B and C drop A from their listing.

Ideally, we can leave `A` and other merged PRs in the navigation comment when we restack and submit B and C.

Blocks #314
Blocks #31

PR #508, opened 2024-12-01
Display merged PRs in navigation comments

VenelinMartinov:
This change adds merged PRs to navigation comments. To do that git-spice now keeps track of merged PRs in the branch state of the bottom branch in each stack. This then gets propagated up to all branches directly above it once the local branch is deleted because of the corresponding PR being merged.

This change is mostly functional but not finished. I wanted to get some feedback if this is the correct approach for this. I've opted to keep a history of merged PRs in the branch state instead of maintaining branches for deleted PRs. It seems like a much simpler change to make and maintaining the branch information does not seem warranted.

This should still be considered a draft because:
- I've not handled many edge cases (ex. what should happen with the PR history if a branch gets inserted at the bottom of a stack)
- There's a bit of refactoring to do (`changeID`s are passed around as strings instead of `forge.ChangeIDs`) 
- I should also add additional tests (any suggestions for known tricky cases would be appreciated!).

fixes https://github.com/abhinav/git-spice/issues/331


abhinav:
Thanks for the PR, @VenelinMartinov. I'll need a few days before I can properly look at this.

The idea of tracking information about merged PRs in their upstacks matches what I was thinking here.
I need to mull over the architectural impact a bit more; e.g. I'm unsure about tracking just the change ID.
I'll try to give this a closer look soon.

VenelinMartinov:
Thank you @abhinav, much appreciated!

testdata/script/stack_submit_multiple_merged_history.txt
```diff
@@ -0,0 +1,235 @@
...
+            - #3
+
+    <sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>
+    <!-- gs:navigation comment -->
+- change: 3
+  body: |
+    This change is part of the following stack:
+
+    - #2
+        - #3 ◀
```
VenelinMartinov, line 227: This one is wrong - the first PR is lost in the history if both 1 and 2 are detected as merged in the same sync.

abhinav:
I didn't want to track the `String()` representation of ChangeIDs because it's lossy.
I've changed it to track the JSON representation of the original ChangeID structure so that the underlying logic can do something richer with it in the future if it needs to (e.g. generate full links `[#123](https://...)` instead of just `#123`).
With that in place, the graph rendering logic doesn't need specialization for merge downstacks: they're just another node in the graph structure from which the comments are generated.

abhinav:
Thanks for your patience @VenelinMartinov. I'm iterating more on this as time permits.
There are a couple design TODOs I have in mind with this that I want to address:

- `branch delete` doesn't feel like the right place for this. I might try to make `repo sync` record this information directly in the upstacks, and then run `branch delete` on the merged branches. But if that turns out to be unfeasible, delete will have to do.
- For `branch onto`, I think it's not right to combine the downstacks. If I had `main -> A -> B` and `main -> C -> D`, and I merge `C` and move `D` onto `A` (`main -> A -> {B, D}`), it's not correct for `C` to show downstack for any of these branches. It was downstack from D in the past, but it was never downstack from B and it's incorrect to say so. I think it's okay to say that with `branch onto`, you're removing this branch from the stack that it used to be part of.

VenelinMartinov:
Thanks for your help @abhinav and sorry for leaving the tests in a broken state - I didn't get the time to revisit these yet.

For `branch onto` I was trying to fix a few cases where editing the stack might lead to an incorrect downstack history.

`stack edit` - this one seems like it should preserve the history - the stack is the same but the order of branches is changed. 

`branch onto` is used to insert a branch at the bottom of a stack: `main -> A -> B -> C` becomes `main -> D -> A -> B -> C`. The fact that the history is saved on the bottom-most branch in the stack makes this case special compared to `main -> A -> D -> B -> C`. The intention here seems the same (add a branch to the stack).
Now if that branch `D` has history, it seems like it should be preserved too and the two combined.
If the branch `D` is based off other branches, then maybe yeah, this is now a different stack so we can lose the history.
I'm also happy to revert the changes which make the first case work. Perhaps inserting a branch at the bottom of the stack is a rare enough use case that it doesn't need to be handled. Please, let me know what you think once you have some time!

abhinav:
For `stack edit`, I wonder if we might be able to special-case it there: if one of the branches being edited has a merged downstack history, transfer it to the first branch as part of, or after the first Onto operation that runs for it (in spice.Service.StackEdit).

> I'm also happy to revert the changes which make the first case work
> Perhaps inserting a branch at the bottom of the stack is a rare enough use case that it doesn't need to be handled

Yeah, that makes sense. (If we go this route, we can defer `stack edit` as well.)
I also think it doesn't have to be perfect about stack history transfer from day one as long as it works for the 80% case:
we'll probably end up adding history transfer logic as cases arise. For example, just thinking out loud: `branch create --insert` at the bottom is another case where the state would drop down, but we shouldn't implement that yet.

If you do revert, I would suggest keeping the test scenarios around and just adjusting the expectations so when that logic is added in the future, there's something already there to work with.

abhinav:
@VenelinMartinov I iterated on this a bit more and was able to address the TODOs I listed above.
Notably: merged downstack propagation happens from `repo sync`, so `branch delete` didn't get modified after all.

I've made the following additional changes:

- `stack edit` will transfer history to bottom-most PR in the stack
- `branch create --below` will transfer history to new bottom-most PR if invoked from bottom of stack

VenelinMartinov:
@abhinav thanks for pushing that to completion!

abhinav:
heh, excellent timing, @VenelinMartinov. I'm in the process of publishing a release.  https://github.com/abhinav/git-spice/pull/539

abhinav:
Release published! homebrew-core update: https://github.com/Homebrew/homebrew-core/pull/203023

PR #511, opened 2024-12-04
git/UpdateTree: Don't leave empty trees behind

abhinav:
When all files from a tree are deleted,
delete the subtree from the parent tree
instead of leaving an empty entry behind.
This matches git's behavior.


abhinav:

- #511 ◀

PR #520, opened 2024-12-20
internal: Generalize topological sort

abhinav:
Extract the topological sort logic used in branch_delete
into a separate package so that we can reuse it in #508.

A small user-facing change of this is that 'branch delete'
will delete branches in a more deterministic order.


abhinav:

- #520 ◀

PR #522, opened 2024-12-21
state: Add 'version' file

abhinav:
There's a possibility that we make breaking changes
to the state store in the future.

To detect and upgrade it automatically, add a 'version' file
to the state store.

Its current value is "1",
and that's what we'll assume when the file is absent.

The binary will refuse to run against a store with a version
newer than the current binary.


abhinav:

- #522 ◀

PR #523, opened 2024-12-21
refac(storage); simplify MemBackend

abhinav:
Split MemBackend into MapBackend and SyncBackend.
Makes it easy to write test cases where a storage backend is pre-filled
with some files.


abhinav:

- #523 ◀

PR #524, opened 2024-12-21
submit: Add --update-only flag

### LINKED ISSUE #516: submit: update-only flag
For commands that submit CRs we should add a flag "upate-only" (or similar) that will not submit any new CRs, only update those that are already open.

This is useful when you have submitted say, A->B, and are currently working on C. 
If you make updates to A/B, you should be able to do `gs stack submit --update-only` without submitting C or anything above it.

abhinav:
For all submit commands, add an `--update-only` flag
that will update existing open CRs but not create new ones.

Resolves #516


abhinav:

- #524 ◀

abhinav:
docs

image

image

PR #526, opened 2024-12-21
repo sync/downstack history: Don't nuke merged branch history

abhinav:
When bubbling up downstack history,
don't nuke the history of a merged branch until it's deleted.
If this branch fails to be deleted/untracked for any reason,
we'll still have this history to work with.

PR #525, opened 2024-12-21
DESIGN: Add notes on merged downstack history

abhinav:
Explain a little of how #508 works in DESIGN for future reference.

PR #527, opened 2024-12-21
repo sync: don't propagate downstack history for unsubmitted branches

abhinav:
Add a test for how merge history propagation behaves
for upstack branches that are unsubmitted.

Suppose I have `feat1 -> feat2`:

- feat1 is ready and I submit a CR
- feat2 is not yet ready for review, so it is unsubmitted
- while I'm still working on feat2, feat1 gets approved and I merge it
- I then submit feat2

feat2 still shows feat1 in its downstack history.

We can change this behavior in the future if it makes sense
but right now I lean towards more information over less.


abhinav:

- #525
    - #526
        - #527 ◀

PR #534, opened 2024-12-27
uitest: Add system for unit testing widgets

abhinav:
Adds a testscript-based system for testing terminal widgets.
This is a generalized, in-memory, Windows-friendly version of with-term
intended for use in unit tests instead of end-to-end tests.
It uses testscript instead of a the custom commands format.

With this in place, we're able to unit test a bunch of the terminal
interactions that were previously untested,
or became untested following RobotView in #480.


abhinav:

- #534 ◀

PR #555, opened 2025-01-24
tests: Handle Git output changes

abhinav:
Looks like `git log --graph --decorate` now includes origin/HEAD
in the output in newer versions.

Rather than try to make the output match different versions,
freeze some of the bits, e.g. never list origin/HEAD.


abhinav:

- #555 ◀

PR #564, opened 2025-02-03
feat(branch): add `gs branch squash`

### LINKED ISSUE #558: New command: `gs branch squash` (same as Graphite squash)
It would be great to have a command like Graphite CLI's [`gt squash`](https://graphite.dev/docs/command-reference#gt-squash)

> Squash all commits in the current branch into a single commit and restack upstack branches.

I was able to make a non-robust shell script that leverages `gs up/down`, to compute the commit range, then use `git` and other unix commands to rewrite the history.

I'd be happy to contribute a PR, esp. if it would involve connecting pre-built APIs that are already in the code (I'm not a go expert, but am interested in learning). Thank you!

```sh
#!/bin/bash

set -e
set -u
set -o pipefail

main() {
	# Get the base branch we branched from
	PREV_BRANCH=$(gs down --dry-run | tail -n 1)

	NEXT_BRANCH_OUTPUT=$(gs up --dry-run)
	if [[ $NEXT_BRANCH_OUTPUT == *"no branches found upstack"* ]]; then
		# No next branch, use HEAD as boundary
		BOUNDARY=$(git rev-parse HEAD)
	else
		# Extract the last line which should be the branch name, ignoring any warnings
		NEXT_BRANCH=$(echo "$NEXT_BRANCH_OUTPUT" | tail -n 1)
		BOUNDARY=$(git rev-parse $NEXT_BRANCH^)
	fi

	# Store the first commit's message in the current branch
	FIRST_MSG=$(git log --format=%B --reverse HEAD...$PREV_BRANCH | tail -n +1 | head -n 1)

	git reset --soft $PREV_BRANCH
	git commit -m "$FIRST_MSG"
}

main "$@"
```

jonboiser:
Resolves #558, using the procedure discussed, namely

1. Verifies that the current branch can be squashed
1. Checks out the current branch in a detached state
2. Resets the detached branch to the base branch
3. Commits all staged changes, using either the original first commit msg, or a provided message
4. Updates the HEAD ref of the current branch to the hash that was created in the previous step
5. Checks out the current branch

### DIFF
--- a/branch_checkout.go
+++ b/branch_checkout.go
@@ -22,7 +22,7 @@ type branchCheckoutCmd struct {
 	checkoutOptions
 
 	Untracked bool   `short:"u" config:"branchCheckout.showUntracked" help:"Show untracked branches if one isn't supplied"`
-	Branch    string `arg:"" optional:"" help:"Name of the branch to delete" predictor:"branches"`
+	Branch    string `arg:"" optional:"" help:"Name of the branch to checkout" predictor:"branches"`
 }
 
 func (*branchCheckoutCmd) Help() string {
--- a/.changes/unreleased/Added-20250202-192729.yaml
+++ b/.changes/unreleased/Added-20250202-192729.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'branch squash: Squashses all commits in the current branch into a single commit and restacks the upstack branches.'
+time: 2025-02-02T19:27:29.806629-08:00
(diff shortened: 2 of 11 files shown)


branch_checkout.go
```diff
@@ -22,7 +22,7 @@ type branchCheckoutCmd struct {
 	checkoutOptions
 
 	Untracked bool   `short:"u" config:"branchCheckout.showUntracked" help:"Show untracked branches if one isn't supplied"`
-	Branch    string `arg:"" optional:"" help:"Name of the branch to delete" predictor:"branches"`
+	Branch    string `arg:"" optional:"" help:"Name of the branch to checkout" predictor:"branches"`
```
abhinav, line 25: 😅  thanks!

abhinav:
Thanks, @jonboiser. 
I made a few follow up edits—mostly invisible stuff but with one significant behavioral change:
Instead of using the first/last commit as the commit message when one isn't provided, this will use the `--template` flag of git commit to fill one up with all of the messages of all of the commits being squashed, and the user will be able to edit that. So the options for a user are: provide a commit message with `-m`, or write one in an editor, with the old ones provided.

This should be good to merge as soon as the build passes.
Thanks for contributing!

PR #570, opened 2025-02-12
doc: Add 'unreleased' support to CLI commands

abhinav:
For new commands, report which version they were released in
(if at all).


abhinav:

- #570 ◀

PR #573, opened 2025-02-15
all(test): prefer t.Context where possible

abhinav:
t.Context was added in Go 1.24 to provide a context
that is valid only for the lifetime of a test function.

This change updates test functions to use this context where possible.

Cases where it can't be used:

- rapid tests
- t.Cleanup functions


abhinav:

- #573 ◀

PR #579, opened 2025-02-19
feat(branch): add sort option for branch list

### LINKED ISSUE #578: Feature: Optionally sort branches by committerdate
When viewing branches with regular git commands, I often use an alias for `git branch --sort=committerdate`. I find it easier to remember what work I had in progress if branches are sorted by recency.

When using the interactive UI of `gs branch checkout`, the branches are sorted lexicographically. I like the TUI of `branch checkout`, but the friction of not having branches sorted by recency means I often switch to non-interactive git commands. This overhead is more noticeable when I have many branches in progress.

It would be great if `gs branch checkout` supported a `--sort` flag with `committerdate` as an option, and a config field equivalent to git's `branch.sort`.

Is this a change you'd be open to? I'd be happy to work on it if so.

schallert:
Fixes #578.

### DIFF
--- a/branch_checkout.go
+++ b/branch_checkout.go
@@ -21,7 +21,7 @@ type checkoutOptions struct {
 type branchCheckoutCmd struct {
 	checkoutOptions
 
-	Sort      string `short:"s" optional:"" config:"branchCheckout.sort" help:"Sort branches by the given field"`
+	Sort      string `config:"branchCheckout.sort" help:"Sort branches by the given field"`
 	Untracked bool   `short:"u" config:"branchCheckout.showUntracked" help:"Show untracked branches if one isn't supplied"`
 	Branch    string `arg:"" optional:"" help:"Name of the branch to checkout" predictor:"branches"`
 }
--- a/branch_checkout.go
+++ b/branch_checkout.go
@@ -30,7 +30,14 @@ func (*branchCheckoutCmd) Help() string {
 	return text.Dedent(`
 		A prompt will allow selecting between tracked branches.
 		Provide a branch name as an argument to skip the prompt.
+
 		Use -u/--untracked to show untracked branches in the prompt.
+
+		Use --sort=<field> to sort branches at the same level by the given field.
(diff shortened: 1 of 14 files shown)


branch_checkout.go
```diff
@@ -21,6 +21,7 @@ type checkoutOptions struct {
 type branchCheckoutCmd struct {
 	checkoutOptions
 
+	Sort      string `short:"s" optional:"" config:"branchCheckout.sort" help:"Sort branches by the given field"`
```
abhinav, line 24: minor: I'd like to keep the short form '-s' free as there's a good chance something that's used more frequently than sort on the CLI will want it. (Also, optional is the default for flags so we can drop it.)
```suggestion
	Sort      string `config:"branchCheckout.sort" help:"Sort branches by the given field"`
```

internal/git/branch.go
```diff
@@ -20,9 +20,26 @@ type LocalBranch struct {
...
+	if opts != nil {
+		sort = opts.Sort
+	}
+
+	var cmd *gitCmd
+	if sort != "" {
+		cmd = r.gitCmd(ctx, "branch", "--sort="+sort)
+	} else {
+		cmd = r.gitCmd(ctx, "branch")
+	}
```
abhinav, line 41: nit: We can compile the list of arguments instead:

```suggestion
	args := []string{"branch"}
	if opts.Sort != "" {
		args = append(args, "--sort="+opts.Sort)
	}
	cmd := r.gitCmd(ctx, args...)
```

internal/git/branch.go
```diff
@@ -65,9 +82,11 @@ func (r *Repository) LocalBranches(ctx context.Context) ([]LocalBranch, error) {
 		return nil, fmt.Errorf("git branch: %w", err)
 	}
 
-	slices.SortFunc(branches, func(a, b LocalBranch) int {
-		return strings.Compare(a.Name, b.Name)
-	})
+	if sort == "" {
```
abhinav, line 85: With the change above, this becomes:
```suggestion
	if opts.Sort == "" {
```

PR #582, opened 2025-02-20
fix(branch onto): rebase loop on conflicts

abhinav:
Fixes an issue reported by a user where the following steps cause
a rebase conflict loop:

- branches A and B share a base, or the base for A is downstack from B
- branch A is moved on top of B, and the operation runs into a conflict
- the user resolves the conflict and runs `gs rebase continue`
- they are asked to resolve more conflicts, as if replaying all commits

The cause of this issue is this:
when the operation to move A onto B runs into a conflict,
git-spice's internal state is not updated to reflect the new base
(this is correct and desirable).

However, when the operation is re-run after resolving the conflict,
it attempts to replay commits `Base..A` onto B,
except `Base..B` overlaps with that range with the new state of the repository.
This forces the user to resolve a bunch of conflicts from replaying those operations,
but the `rebase --continue` does not ever succeed, and the state is never updated.
so it keeps retrying the same commits on top of B over and over again.

The fix for this is fairly straightforward:
if B is in the Base..A commit range, rebase only B..A instead of Base..A.


abhinav:

- #582 ◀

PR #584, opened 2025-02-21
fix(upstack onto): disable upstack branches in prompt

abhinav:
When upstack onto presents a prompt to select the new base,
it should not allow selecting the current branch or any of its upstacks
as that operation will fail.


abhinav:

- #584 ◀

PR #585, opened 2025-02-21
fix(git): show command name in log prefix

abhinav:
As we were injecting configurations into git commands at runtime,
log prefixes for affected commands were just `-c`.

To fix this, make newGitCmd aware of extraConfig
so it can guess at a command name while ignoring the `-c` flags,
and add the ability to change the log prefix,
as that's required for rebase command when --no-edit is enabled.


abhinav:

- #585 ◀

PR #587, opened 2025-02-21
chore(gittest): Use testscript.Params.Files

abhinav:
testscript.Params.Files support was added in a prior version,
obviating the need to set up a temporary directory with just the script.
We still use it to run scripts from in-memory strings,
but otherwise, we can rely on the Files parameter for the work here.


abhinav:

- #587 ◀

PR #589, opened 2025-02-22
chore(kong): BindToProvider singletons workaround

abhinav:
BindToProvider in Kong will create new instances of bindings
that come from a provider function every time they're requested
by another binding.

This means that we always re-open some of these resources
multiple times: e.g. repository is opened when requested by state.Store,
by spice.Service, and then by individual commands.

Work around this with a sync.Once until the new
[BindSingletonProvider option](https://github.com/alecthomas/kong/pull/501)
is released.

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,7 +8,9 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"reflect"
 	"strconv"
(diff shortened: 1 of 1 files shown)


abhinav:

- #589 ◀

PR #590, opened 2025-02-22
branch prompt: Untracked branches are based on trunk

### LINKED ISSUE #443: branch widget: untracked branches should be on trunk
The `gs bco` branch widget shows untracked branches as their own root nodes.
They should *probably* be based off trunk instead of being root nodes. Probably.
Opening this issue to consider this idea and then keep/discard.

abhinav:
We currently list untracked branches as root nodes
when presenting the branch selection tree.

This doesn't seem right: untracked branches are still based off of trunk
(unless they have completely divergent histories,
which we won't account for -- at least for now).

Present untracked branches as being based on trunk.

Resolves #443


abhinav:

- #590 ◀

PR #601, opened 2025-02-24
version: --short flag to print only the version

### LINKED ISSUE #572: Add short version flag
I use Chezmoi, and use version outputs [1] to know when to re-generate shell completions. Unfortunately, it only works with single lines, and the current `--version` outputs multiple lines.

I'd like to add a short version flag, and if you're open to it, I'm happy to submit a PR! From looking at the code, I was thinking of adding either an extra flag (`--short`) or to change the default behaviour to be just the version, and then if `--verbose` is on, then to include all the current output.

[1] https://www.chezmoi.io/reference/templates/functions/output/

abhinav:
Per discussion in #572, add a --short flag to the version command
to print only the version number.

[skip changelog]: version command is unreleased,
this is just an addition to that.

Resolves #572

### DIFF
--- a/doc/includes/cli-reference.md
+++ b/doc/includes/cli-reference.md
@@ -1034,3 +1034,7 @@ gs version [flags]
 
 Print version information and quit
 
+**Flags**
+
+* `--short`: Print only the version number.
+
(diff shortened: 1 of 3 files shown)


abhinav:

- #601 ◀

jesse-c:
Just wanted to say thank you again for this! I'm using it [now](https://github.com/jesse-c/dotfiles/commit/c89bda06ec9136e7081915b1068214f5fa5044c7).

PR #604, opened 2025-02-25
internal/spice: Drop Forge from Service

abhinav:
Service doesn't need to know Forge (at least at this time)
as all logic is implemented outside it.

[skip changelog]: no user-facing changes

PR #605, opened 2025-02-26
refac(forge): extract global registry into type

abhinav:
Move the logic from the global forge registry into a Registry type
that mirrors the APIs (more or less).
In this change, we still have a global registry;
following this we can drop the global registry.


abhinav:

- #604
    - #605 ◀
        - #606

PR #613, opened 2025-03-06
fix(gitlab): Actually support GITLAB_TOKEN

### LINKED ISSUE #612: Support self hosted gitlab for `gs stack submit`
I see that supported remotes currently are publicly available Gitlab and Github.
Though I dont see a strong reason to allow git-spice to be able to push to self-hosted Gitlab instances for following reasons:
- All the api calls remains the same
- Git push commands dont change for a tracked repo with Origin configued
- Even if the support for Merge request wont be too challenging, the tool can sync the changes easily

I guess the tool can have additional config to support it fully like
GITLAB_HOST
GITLAB_TOKEN

abhinav:
The GitLab forge was rejecting use of the GITLAB_TOKEN env var
for authentication despite advertising support for it.

The issue was that the client constructor was exploding
when the auth type was `AuthTypeEnvironmentVariable`
because it was not in the list of known auth types.

Fix this by adding a case for it, and add tests to verify.

Resolves #612

### DIFF
--- a/.changes/unreleased/Fixed-20250305-215537.yaml
+++ b/.changes/unreleased/Fixed-20250305-215537.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'gitlab: Fix rejection of `GITLAB_TOKEN` environment variable for authentication.'
+time: 2025-03-05T21:55:37.669148-08:00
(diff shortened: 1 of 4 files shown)


abhinav:

- #613 ◀

Issue #610, opened 2025-03-03
Enhancement to `gs log`: An option to display URL to Change Request


jonboiser:
Currently, `gs log` has this output:

```
❯ gs ls
┏━■ branch (#5092) ◀
```

Which only has, in this case, the GitHub PR number.

Would it be possible to have a flag, or have another `log` subcommand that shows more information like the URL, so I could Cmd+Click it to open it from the terminal?

I'd be happy to create a PR for this. It seems like other features like `gs branch submit` access this URL from the local database.

abhinav:
Hello, yes, a configuration option to do this would be welcome.
It can be one of the config-only fields on `gs log`.
I don't have an obvious name suggestion for it, but it's probably in the category of `spice.log.*`.
Perhaps `crFormat={id | url}` with `id` being the default.

PR #615, opened 2025-03-09
feat: add `spice.log.crFormat` configuration

jonboiser:
Implements #610 .

I copied the pattern of `branch_submit.go`, which gets the remote CR URL after it updates a branch:

- injects `secretStash` and ` forges` into the command
- resolves the remoteRepository if `spice.log.crFormat = 'url'`, and prints the URL of any submitted branch

I noticed that `gs log` has some latency with the config turned on. Is there a way to make this more optimized? I had thought originally that `gs` stored a local copy of the branches, or that `gs` would have used locally-available information like `git remote` and the forges' URL schemes to create this URL.


abhinav:
> I noticed that gs log has some latency with the config turned on. Is there a way to make this more optimized? 

Yeah, right now this is querying the remote forge on demand.
This information is available in the store, but it looks like the Forge ChangeMetadata API needs to be adjusted a tad bit.
Let me take a poke at it; I'll update your PR.

abhinav:
@jonboiser I added #617 for a cheap, local-only way to calculate the CR URL and integrated that here. Try it out, please?
I'll merge this PR afterwards if all looks good on your end.
Thanks for the change!

jonboiser:
@abhinav Tested it out and it looks good. `gs ls` is equally fast on both configurations. Thanks for the help!

abhinav:
Excellent! Thanks for the contribution.

PR #617, opened 2025-03-09
forge: Learn to calculate change URLs

abhinav:
Needed for #610, #615

We currently don't expose a standard way to determine the URL
at which a CR can be viewed in a forge-agnostic way.

This adds a ChangeURL method to Forge that can be used for this purpose.

The method is a bit of a hack as this information should typically
come from the Repository, except right now,
we do not have a means of narrowing down to a Repository
without "opening" it, which may perform network requests.

There's an in-flight change that separates the act of
parsing a repository URL and opening it,
which will allow making this cleaner in the future.


abhinav:

- #617 ◀

PR #606, opened 2025-02-26
refac(forge): Drop global registry

abhinav:
Replace the global forge registry with
an explicitly plumbed Registry object.

[skip changelog]: no user facing changes

PR #624, opened 2025-03-15
refac(forge): Split parsing URL and opening Repository

abhinav:
Right now, the act of parsing a remote URL for a Forge's repository
has to be repeated a bunch because we don't hold onto that information.
There is no first-class reference to information we have obtained
about a repository after calling `MatchURL` or similar,
so we end up doing repeatedly as needed.

This was especially evident in #615, where we had to add a ChangeURL
method to the Forge, which would re-parse the remote URL yet again
just to be able to generate the full URL for a change.

This refactors the relationship between Forge and Repository a tad-bit
to introduce a first-class reference to information parsed
from a Git remote URL.

We introduce a new RepositoryID type which represents this information.
In case of GitHub and GitLab, this is `(url, owner, repo)`.

This allows moving ChangeURL as a method onto RepositoryID
so that the Forge can just return the RepositoryID on match,
and the caller can use the RepositoryID to generate the change URL.


abhinav:

- #604
    - #605
        - #606
            - #624 ◀

Issue #619, opened 2025-03-12
log: show "pushed" status


abhinav:
### Discussed in https://github.com/abhinav/git-spice/discussions/618

<div type='discussions-op-text'>

<sup>Originally posted by **fahmifan** March 11, 2025</sup>
Is there a way to show the `pushed` status of a branch in the log? 
I think this is helpful when I'm using non github/gitlab repo, 
This is a feature in git-stack that I found usefull
![image](https://github.com/user-attachments/assets/6d790f1a-f7cc-44e5-a9ac-ece4c2b3ed70)

I changed to git-spice because the restack is better</div>

jonboiser:
This has been bothering me recently too, not knowing which code has been pushed to the remote.

I have something that seems to work on my local build, similar to what we did in #610 . Can I submit a PR for this?

abhinav:
@jonboiser Absolutely! I had a WIP comment giving guidance on what would be needed here; I hadn't finished writing it.
I think you have a pretty good idea about the relevant pieces already, so please feel free to post a PR; I'm happy to iterate on it with you to get it merge-ready.

I'm thinking that the config is probably `spice.log.pushed` or `spice.log.pushedFormat`. (Opinions welcome!)
Possible values: `true`/`false` with a default of true.

Future expansion: 
I'm imagining that we would add a possible value of `aheadBehind` in the future to show divergence similarly to a bunch of popular Git shell prompt plugins ([example](https://github.com/magicmonty/bash-git-prompt?tab=readme-ov-file#examples)).

PR #625, opened 2025-03-16
feat(log): adds `log.pushedFormat` config

jonboiser:
Implements #619:

- The config is `spice.log.pushedFormat = true | false` (default: true)
- In `gs log`, the indicator shows up in green text (IDK if this is great for all terminal themes, though. Does the styling library ensure decent contrast)?
- Since this updates the default behavior of log, many of the existing test scripts need to be updated.

```
❯ ./bin/gs ls
┏━■ feat-log-pushed-style (https://github.com/jonboiser/git-spice/pull/5) (pushed) ◀
main
```


abhinav:
Thanks! 

> the indicator shows up in green text

IMO that clashes with the "needs restack" text.
I think I'm going to change that to a dim yellow, e.g.

image

(If you don't like that, I'm down to also include a means for changing the color for this.)

> Since this updates the default behavior of log, many of the existing test scripts need to be updated.

I did not think of that. I realize now that it'll be pretty noisy since it'll highlight the majority case ("things are up to date").
I'm going to flip this around to report "needs push" similar to "needs restack" so that it's less noisy.

image

abhinav (approved):
Thanks, @jonboiser.
I've got a change incoming with aheadBehind support.
(I suspect that's what I'll be using in my day-to-day for this.)

PR #641, opened 2025-04-06
Support building with cpu, mem, and trace profiles

abhinav:
'mr build --pprof` will generate a build
that will add three new flags: --cpuprofile, --memprofile, and --trace.
These can be used to generate profiles and debug performance issues.

These are not included by default to avoid bloating the binary size.


abhinav:

- #641 ◀

Issue #656, opened 2025-05-13
Add spice.branchCreate.prefix config (or something similar)


klnusbaum:
Graphite has a [configuration option](https://graphite.dev/docs/configure-cli#branch-naming-settings) where, when creating a new branch, you can specify a prefix that all branches should be created with. One example of when this is useful: if you're working in a large repo, it's nice to be able to prefix your branches with your name/initials/handle.

I looked through the configuration docs and couldn't find something analogous. I'm wondering if we could add a config option called `spice.branchCreate.prefix`. The value of the config key would then be prepended to the what ever name is passed to the `gs branch create` command.

abhinav:
Fixed with #658

PR #658, opened 2025-05-13
Add branch prefix option

klnusbaum:
It'd be nice to be able to always add a prefix to the name of any branch that gets created.

Addresses https://github.com/abhinav/git-spice/issues/656

### DIFF
--- a/branch_create.go
+++ b/branch_create.go
@@ -14,7 +14,7 @@ import (
 )
 
 type branchCreateConfig struct {
-	Prefix string `default:"" config:"branchCreate.prefix" help:"Prepend a prefix to the name of the branch being created"`
+	Prefix string `default:"" config:"branchCreate.prefix" help:"Prepend a prefix to the name of the branch being created" hidden:""`
 }
 
 type branchCreateCmd struct {
--- a/branch_create.go
+++ b/branch_create.go
@@ -13,7 +13,13 @@ import (
 	"go.abhg.dev/gs/internal/text"
 )
 
+type branchCreateConfig struct {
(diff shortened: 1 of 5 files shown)


branch_create.go
```diff
@@ -217,6 +223,8 @@ func (cmd *branchCreateCmd) Run(
 		}
 	}
 
+	branchName := cmd.Prefix + cmd.Name
```
klnusbaum, line 226: I went back and forth here on whether or not we should create a new variable (as I ended up doing) or just to modify `cmd.Name` like we do in the if block above. I ultimately choose this since the above if block sets cmd.Name since it has no value. In this case, we're using the now set value plus a different command argument to create the end branch name that we use.

branch_create.go
```diff
@@ -13,7 +13,13 @@ import (
 	"go.abhg.dev/gs/internal/text"
 )
 
+type branchCreateConfig struct {
```
klnusbaum, line 16: wasn't sure if you wanted this struct to be private or public, or where it should go. Happy to move it.

doc/src/cli/config.md
```diff
@@ -100,6 +100,18 @@ and use the `--commit` flag to commit changes when needed.
 - `true` (default)
 - `false`
 
+### spice.branchCreate.prefix
+
+<!-- gs:version v0.14.0 -->
```
klnusbaum, line 105: used `0.14.0` since that looks like the next version that will get cut.

klnusbaum: looks like this actually causes a build error. Is there an alias to use for unreleased versions?

branch_create.go
abhinav: Private is good for now 👍

abhinav (approved):

Issue #655, opened 2025-05-12
Better workflow for renaming remote branches


abhinav:
gs branch rename only renames the local branch, and leaves the remote alone.
AFAIK there's no way to rename the remote ref without deleting and re-pushing.

Right now, if a user wants to rename the local branch and the remote, the workflow is:

```
git push --delete origin $oldName
gs branch untrack
gs branch track
gs branch submit
```

Without the `untrack`/`track`, the git-spice state will still remember `$oldName` as the remote name for the branch and attempt to push to it.
This behavior was intentional, so as not to accidentally disconnect from PRs after pushing a branch.
However, as evident above, this makes the case where the user _does_ want to rename the branch (maybe because they haven't created a PR) annoying.

Some options:

- in `gs branch rename`, if the branch doesn't have a PR associated, offer to rename the remote
- if we can determine in `gs branch submit` that the remote ref was deleted (maybe since we use `--force-with-lease`), use that knowledge to offer to use the new remote name

PR #662, opened 2025-05-18
state: detect upstream ref deletion

abhinav:
If an upstream branch reference is deleted manually
(e.g. with git push --delete, or by deleting the remote branch),
then the state's local knowledge of the upstream branch name is stale
and should be deleted.

This does not solve, but helps with issues like #655,
as the user won't have to untrack the branch
to delete stale information.


abhinav:

- #662 ◀

PR #666, opened 2025-05-21
feat: add ctrl+k and ctrl+j navigation keys for select uis

loganlinn:
in addition to up and down arrows, to navigate selection lists where text input is allowed to refine the list of options, this adds ctrl+k and ctrl+j as alternatives, similar to other tools like fzf.

i didn't see existing tests for up/down to build off of, but i did manually test this locally and it behaves as expected.


abhinav (approved):
Sure! Thanks for adding.

PR #667, opened 2025-05-22
Custom slog-based Logger

abhinav:
charmbracelet/log has gotten us pretty far,
but for the level of control I want going forward,
we need our own wrapper around slog.

This changes just adds the new logger implementation.
Following changes will switch to it.

logtest and log.Writer are copy-pasted from the versions in logutil.

[skip changelog]: This change contains no user facing changes yet.

### DIFF
--- a/internal/log/attr.go
+++ b/internal/log/attr.go
@@ -1,0 +1,13 @@
+package log
+
+import "log/slog"
+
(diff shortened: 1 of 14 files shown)


abhinav:

- #662
    - #667 ◀

PR #668, opened 2025-05-22
Update all logging to use internal/log

abhinav:
While updating most log sites was just a search-and-replace,
the following changes are manually made:

- Changelog
- Updates to linter rules to enforce loggercheck on internal/log
- Changing dumpmd to include when a flag was added to the CLI
- Updating git/cmd.go and main.go to use the new logger
- Fixed an annoyance with interactive rebase driven by git-spice
  where it tried to redirect the output of the command through the
  logger; this was printing some garbage escape characters at the end.

On the user front, --verbose is now just an alias for --verbosity=1.
The old `-v` may now be supplied multiple times.


abhinav:

- #662
    - #667
        - #668 ◀

PR #674, opened 2025-05-24
chore: use BranchExists instead of PeelToCommit

abhinav:
In cases where PeelToCommit is only used to check if a branch exists,
use BranchExists instead.


abhinav:

- #674 ◀

PR #676, opened 2025-05-24
fix(branch create): prefix in branch name conflict resolution

abhinav:
When 'branch create' generates a branch name based on the commit,
if the branch name already exists, it appends a number to the name.
Right now, we add the prefix to the generated name afterwards,
which means that there's a chance that the final name
will conflict with an existing branch.

This change fixes this by ensuring that the prefix is taken into account
when generating the branch name.

[skip changelog]: change hasn't been released yet

### DIFF
--- a/branch_create.go
+++ b/branch_create.go
@@ -170,7 +170,10 @@ func (cmd *branchCreateCmd) Run(
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


abhinav:

- #674
    - #676 ◀

PR #702, opened 2025-06-21
config: add spice.{logShort, logLong}.crFormat options

### LINKED ISSUE #693: `gs ll`/`gs ls` + `spice.log.crformat=url`
I have currently set `spice.log.crformat=url` to easily be able to click and open the PRs in the stack in my browser. Having the links present at all times _do_ take up quite a bit of space however, and I do not always need the information.

It would be convenient if I could either:
- have separate `spice.log.crformat` options for `gs ll` and `gs ls` (my preference), or
- be able to say if I want an _extra_ short view to appear, e.g.: 
    - `gs ls --compact` / `gs lsc` (`compact (c)`)

I think the latter can potentially be a bit confusing (i.e. asking for `long` but also `compact`), so if the former is / can be possible, that'd be amazing!

abhinav:
Adds the `gs ll` and `gs ls` specific overrides for crFormat
discussed in #693.

Resolves #693

### DIFF
--- a/.changes/unreleased/Added-20250620-211723.yaml
+++ b/.changes/unreleased/Added-20250620-211723.yaml
@@ -1,0 +1,5 @@
+kind: Added
+body: >-
+  log short: Add `spice.logShort.crFormat` configuration option.
+  This takes precedence over `spice.log.crFormat` for `gs log short`/`gs ls`.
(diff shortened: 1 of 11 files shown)


abhinav:

- #702 ◀

Issue #700, opened 2025-06-17
Guardrails for `gs ca`


meyer:
I’m very much in the habit of using `gs ca -a --no-edit` to amend my changes when i’m working on a stack. I make the following mistakes fairly frequently:
- I accidentally amend when I’m on main instead of `gs bc -a name-of-branch`
- I accidentally amend when restacking instead of adding my changes and using `gs rbc`

I’m in the zone, my brain is on autopilot, and i accidentally amend instead of git add or create a branch.

it would be kinda neat if there were guardrails for these cases, maybe guardrails that can be turned off for those sickos who want to amend main or amend the wrong commit when rebasing. I’m thinking something like a prompt:
- "looks like you’re on main, are you sure you want to amend? do you want to create a new branch instead?" [create branch] [exit without amending] [amend]
- "looks like you’re in the middle of a rebase. do you want to add your changes instead of amending?" [add changes] [exit without amending] [amend]

abhinav:
@meyer I had some time to work on this this weekend. #709 and #710 implement the above.

While trunk was straightforward, discouraging amend during a rebase is not always desirable.
You may be using an `edit` instruction to amend an existing commit.
I think the only time we want to discourage commit amend, and push for `git add` is if the rebase occurred due to a conflict.

meyer:
@abhinav this is so sick! thanks so much for getting this in!!

PR #709, opened 2025-06-22
commit amend: prompt to create a branch on trunk

abhinav:
If 'commit amend' is called on the trunk branch,
prompt the user to create a branch instead
as that's rarely what users will actually want to do.

Refs #700


abhinav:

- #709 ◀

PR #713, opened 2025-06-25
feat: support shell commands for custom shorthands

### LINKED ISSUE #385: custom shorthands: Support external commands
In the style of git aliases, we should have support for shorthands to be shell commands if prefixed by `!`, e.g.

    git config spice.shorthand.foo '!gs br && gs bs'

abhinav:
Implement support for shell commands in custom shorthands
by prefixing commands with '!'.
This allows users to define aliases that execute arbitrary shell
commands through the gs command interface.

Shell command aliases receive arguments as positional parameters
and can use shell parameter expansion for flexible command construction.
The implementation includes safeguards against infinite recursion.

Resolves #385


abhinav:

- #713 ◀

PR #718, opened 2025-07-05
git/LocalBranches: add support for filtering by pattern

abhinav:
Add support to git/LocalBranches to filter branches by patterns.
This will be used to determine whether a specific branch is checked out
in another worktree.

PR #720, opened 2025-07-05
{repo sync, branch delete}: handle worktrees better

### LINKED ISSUE #665: repo sync: Branch deletion complaint when worktrees are in use
We previously added support to `repo sync` to handle the case when main was checked out in another worktree.
Right now, there's a bug where, if we have:

- wt1: checks out main
- wt2: checks out feature1, with feature2 stacked on top of it

If we run `gs repo sync` from wt2, it will refuse to delete feature1 and check out feature2.

abhinav:
This changes 'repos sync' and 'branch delete' to handle two worktree
cases better:

- repo sync: if a branch slated for deletion is checked out
  in another worktree, skip it
- branch delete: if the branch that would be checked out after
  deletion is checked out in another worktree, detach HEAD

Resolves #665


abhinav:

- #718
    - #719
        - #720 ◀

Issue #568, opened 2025-02-11
allow configuration of stack submit answers


SamSokolin:
I'd like to be able to pre-populate the answers for template, body, etc. w/o being prompted.

davidmc24:
I think those sound like great config options.  In particular, I'm interested in a config option to make draft the default.

abhinav:
#721 should handle at least the draft/no-draft default case seeing as it's a pretty popular request.

abhinav:
Going to mark this as done for now. Please feel free to open a new issue if the existing configuration dials aren't enough.

PR #721, opened 2025-07-05
submit: add 'spice.submit.draft' to change --draft/--no-draft default

abhinav:
Add a spice.submit.draft configuration option
that will control the default value of --draft/--no-draft
when creating new CRs with `gs branch submit` and friends.

In interactive mode, this merely controls the default value
of the prompt that asks for the draft status.
(So a user can just press Enter to accept the default.)

In non-interactive mode, this sets the value that will be used
if the user does not explicitly pass --draft or --no-draft.

Tests include all combinations of:

    { interactive, non-interactive } x { --draft, --no-draft, no flag }

Refs #568


abhinav:

- #721 ◀

PR #723, opened 2025-07-06
submit/web: set to 'created' to open new CRs only

### LINKED ISSUE #649: Feature request: add option to spice.submit.web to open browser tab only on created CRs
Currently, `spice.submit.web` has options `true` and `false`. I'm proposing a third value, something like `create-only`, that only opens a browser tab for CRs that are being created, and skips those that are only being updated.

In my workflow, I create the CRs initially as drafts, then go to the browser Github tab and fill out what I need in the description, assign reviewers if needed, see result of CI checks, etc. For this, setting `spice.submit.web` as `true` is great. However, after that, I don't really need to open the CRs again after each change, unless I get some review comments for example. And even if I did need to see them, most of the time I have my open and active CRs already open in my browser. Thus, I think avoiding the browser tab open in submits after the initial creation of the CR could be useful.

What do you think? Do you see value in an option like that? Thanks for the great tool btw!

abhinav:
The --web flag now accepts an argument,
which can be 'true', 'false', or 'created'.
true and false match today's behavior, while 'created'
means that only newly created CRs will be opened in a browser.

For backwards compatibility, the argument is optional:

    --web    # equivalent to --web=true
    --no-web # equivalent to --web=false

The configuration option `spice.submit.web` also accepts the same values.

Resolves #649

### DIFF
--- a/.changes/unreleased/Changed-20250706-152414.yaml
+++ b/.changes/unreleased/Changed-20250706-152414.yaml
@@ -1,0 +1,6 @@
+kind: Changed
+body: >-
+  submit: The --web flag and the spice.submit.web configuration option
+  may now be set to 'created' to only open new CRs in a browser.
+  Updates to existing CRs will not be opened in a browser.
+time: 2025-07-06T15:24:14.921298-07:00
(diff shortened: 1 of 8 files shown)


abhinav:

- #722
    - #723 ◀

PR #726, opened 2025-07-08
[0.15] fix(submit): push should run from worktree

### LINKED ISSUE #725: Pre-push hooks that check worktree contents
Our repository has a pre-push hook that, in effect, runs a `make` target in the repository to lint code and run tests. With the recent changes to support worktrees (I think), `gs branch submit` now runs these from within the `.git` directory. [Here](https://github.com/abhinav/git-spice/blob/f9ead3879aa52f21732721d9e52bd59e9437ef63/branch_submit.go#L454) it calls `repo.Push()`.

At first I thought "Oh OK, our hook needs to step out of `.git`" but then that would not work right with worktrees – we'd be running the checks against the main checkout, not the worktree where I made the changes.

How should this work? Are we holding it wrong by assuming that there is a checkout of the repository available at all? Or do git hooks legitimately make `git push` a worktree-dependent operation?

abhinav:
'git push' should be run from the worktree where the submit was run
instead of the .git directory of the repository
even if the push operation can be done without a worktree
as pre-push hooks may depend on the worktree state.

Resolves #725


abhinav:

- #726 ◀

abhinav:
https://github.com/abhinav/git-spice/actions/runs/16132290199

PR #729, opened 2025-07-09
feat: repo restack command

### LINKED ISSUE #505: Add `repo restack`
After something is merged to `main`, I tend to check out the trunk and run `stack restack` to update all stacks at once and keep them fresh.  

It would be nice if we could restack everything without needing to check out the trunk itself, allowing us to continue working from where we left off after the restack.

abhinav:
Add a 'repo restack' command that restacks all tracked branches.
Branches have to be sorted in topological order
based on their base relationships before being restacked.

Resolves #505


abhinav:

- #729 ◀

PR #736, opened 2025-07-12
fix: concurrent Git storage usage, bad base hash after manual rebase

### LINKED ISSUE #731: Rebasing with `--update-refs`
I recently hosted a workshop on your tool (thanks again for creating this!) and found the following warning message when comparing git-spice to `git-rebase <branch> --update-refs`

## To reproduce
go to the trunk of a git repository and write:

```
gs bc A -m "A"
gs bc B -m "B"
gs bc C -m "C"
gs bc D -m "D"
```
go to the bottom of the stack (branch `A`) and make a new commit
```
gs bco A
git commit --allow-empty -m "new commit"
```
go to top of stack then rebase using "vanilla" git command with the `--update-refs` flag
```
gs U
git rebase A --update-refs
Successfully rebased and updated refs/heads/D.
Updated the following refs with --update-refs:
        refs/heads/B
        refs/heads/C
```
However, if I, e.g., show the `git-spice` log now, I get the following warning messages:
```
gs ls
WRN could not update ref: retrying  
  error=
  | exit status 128
  | stderr:
  | fatal: update_ref failed for ref 'refs/spice/data': cannot lock ref 'refs/spice/data': is at 24fa0bf06c09eaf552d3ba1e6f8859ce65d8cc0d but expected 6d418fc6ecfa6f46ac48f9d360ab477cdffc414b
WRN could not update ref: retrying  
  error=
  | exit status 128
  | stderr:
  | fatal: update_ref failed for ref 'refs/spice/data': cannot lock ref 'refs/spice/data': is at 24fa0bf06c09eaf552d3ba1e6f8859ce65d8cc0d but expected 6d418fc6ecfa6f46ac48f9d360ab477cdffc414b
WRN could not update ref: retrying  
  error=
  | exit status 128
  | stderr:
  | fatal: update_ref failed for ref 'refs/spice/data': cannot lock ref 'refs/spice/data': is at 09ba494c8b9149dbd2ead00d0888a28ca374c7c7 but expected 24fa0bf06c09eaf552d3ba1e6f8859ce65d8cc0d
```

My understanding is that git-spice still works and was indeed able to resolve the `update_ref` issue. Perhaps it would be useful if an `INF` message was added that states that `retrying` was successful, or something similar?

---

```
gs --version
git-spice 0.15.2
```

abhinav:
This PR fixes a couple issues discovered in trying to fix #731.

First, there's a data race in the Git backend:
when multiple processes try to write to it,
the `update ref` operation can fail.
We accounted for this and retry, but with enough updates,
the operation can actually fail.

To fix this, add locking around the Git backend's read/write operations,
ensuring that only one process can write at a time.
The included test tries to run a bunch of operations on it concurrently,
and fails without the fix.

Secondly, we discovered that in a couple places, including `log long`,
we did `LookupBranch`/`LoadBranches` before calling `VerifyRestacked`.
This means we loaded the branch information (including the base hash)
and then called `VerifyRestacked`, which, despite the read-only-looking name
may change the branch's base hash if the branch was manually restacked
by the user (using `git rebase` instead of `gs branch restack`).
This would result in the operation using the wrong base hash.

The following commands were affected:

- `log long` would show the wrong list of commits
  for the first run after a manual rebase.
- `branch squash` would use the wrong base hash when squashing the commits,
  resulting in an extraneous no-op commit.
- `branch submit` was not negatively affected, but it was updated nonetheless

(Test cases included for `log long` and `branch squash`.)

Resolves #731

### DIFF
--- a/.changes/unreleased/Fixed-20250711-212136.yaml
+++ b/.changes/unreleased/Fixed-20250711-212136.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: Fix concurrent updates to git-spice's state causing "cannot lock ref" warnings.
+time: 2025-07-11T21:21:36.546923-07:00
--- a/.changes/unreleased/Fixed-20250711-212236.yaml
+++ b/.changes/unreleased/Fixed-20250711-212236.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'log long: Fix incorrect list of commits following a manual rebase operation.'
+time: 2025-07-11T21:22:36.27471-07:00
--- a/.changes/unreleased/Fixed-20250711-215056.yaml
+++ b/.changes/unreleased/Fixed-20250711-215056.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'branch squash: Fix no-op commit left behind in some scenarios when ''branch squash'' is run after a manual rebase.'
+time: 2025-07-11T21:50:56.884734-07:00
--- a/branch_squash.go
+++ b/branch_squash.go
@@ -46,17 +46,17 @@ func (cmd *branchSquashCmd) Run(
 		return errors.New("cannot squash the trunk branch")
 	}
 
-	branch, err := svc.LookupBranch(ctx, branchName)
-	if err != nil {
-		return fmt.Errorf("lookup branch %q: %w", branchName, err)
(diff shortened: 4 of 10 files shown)


abhinav:

- #736 ◀

PR #737, opened 2025-07-12
test(github, gitlab): attempt to unflake PAT auth tests

abhinav:
The uitest script here seems to be wrong.
Unclear why we're grabbing the snapshot after sending the keys
instead of before.
I wasn't able to reproduce the flake locally,
but this should fix it.


abhinav:

- #737 ◀

PR #738, opened 2025-07-12
repo init: warn if trunk is behind upstream

abhinav:
If trunk is behind upstream,
but a local branch is made on top of latest upstream,
and then someone onboards onto git-spice,
restack commands will attempt to replay the upstream commits
when the local branch is rebased on top of trunk.

This guards against that by warning the user
when trunk is behind upstream at initialization time.


abhinav:

- #738 ◀

PR #740, opened 2025-07-12
refac(branch checkout): extract checkout logic into a handler

abhinav:
This seems like further reinforcement of the handler idea.
The commands that need to use checkout functionality
(and verify branch is restacked after checkout)
now get this for free without `(&branchCheckoutCmd{...}).Run(...)`
and worrying about the N arguments it takes.


abhinav:

- #739
    - #740 ◀
        - #741

PR #741, opened 2025-07-12
refac(submit): extract into a handler

abhinav:
Building upon the prior two changes,
this refactors the submit commands to all be backed
by the same SubmitHandler implementation.

This allows completely dropping the "submit session" concept
as its only purpose was to pass information between "*stack submit"
and "branch submit" commands.


abhinav:

- #739
    - #740
        - #741 ◀

PR #749, opened 2025-07-14
fix(repo init): validate trunk branch is local

### LINKED ISSUE #734: repo init: don't allow remote refs as trunk
One user was able to use `gs repo init --trunk origin/master`.
This will break things in unexpected ways; we make assumptions about trunk being a local branch.
We should guard against this.

abhinav:
Add validation to ensure that the trunk branch specified with --trunk
is a local branch rather than a remote reference.
When a non-local branch is specified, show a error message and fail.

Resolves #734


abhinav:

- #749 ◀

PR #768, opened 2025-07-22
up: Use branch graph to avoid reloading state

abhinav:
Instead of calling ListAbove in each iteration of the up command,
load the branch graph once and use it to iterate through the branches.

With a 10-level deep stack, the new version is >8x faster.

```
❯ hyperfine --warmup 5 'gs up -n 10' '../main/bin/gs up -n 10'
Benchmark 1: gs up -n 10
  Time (mean ± σ):      3.061 s ±  0.012 s    [User: 9.172 s, System: 10.825 s]
  Range (min … max):    3.048 s …  3.091 s    10 runs

Benchmark 2: ../main/bin/gs up -n 10
  Time (mean ± σ):     347.1 ms ±   3.5 ms    [User: 940.8 ms, System: 1078.6 ms]
  Range (min … max):   343.5 ms … 353.5 ms    10 runs

Summary
  ../main/bin/gs up -n 10 ran
    8.82 ± 0.10 times faster than gs up -n 10
```

### DIFF
--- a/.changes/unreleased/Fixed-20250722-102117.yaml
+++ b/.changes/unreleased/Fixed-20250722-102117.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'up: Don''t re-load state repeatedly when going up >1 branches. This is >8x faster in degenerate cases.'
+time: 2025-07-22T10:21:17.543993-07:00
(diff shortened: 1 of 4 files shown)


abhinav:

- #768 ◀

PR #770, opened 2025-07-23
fix(restack): leverage branch graph to avoid reloading

abhinav:
Avoid loading the branch graph for the downstack/upstack traversal
during a restack operation by taking advantage of the branch graph.
This won't be a major difference versus reloading
as we're just going from two to one reload.


abhinav:

- #769
    - #770 ◀

PR #771, opened 2025-07-23
fix(restack): worktree aware restack

abhinav:
When running stack-scoped operations, if a branch to be restacked
is checked out in another worktree,
skip it instead of failing the command.


abhinav:

- #769
    - #770
        - #771 ◀

Issue #775, opened 2025-07-25
Add ability to configure default "update only" behavior for `gs stack submit`


klnusbaum:
`gs stack submit --update-only` is useful behavior and some users would like the ability to configure git spice such that, by default, all uses of `gs stack submit` have this behavior.

PR #776, opened 2025-07-25
Add 'spice.submit.updateOnly' config to set --update-only by default

klnusbaum:
The `--update-only` flag is useful and some users would like that to be
the default behavior. We can make this behavior configurable. We follow
the pattern currently used by `Draft` option. I.e. we create a new,
hidden, config flag called `UpdateOnlyDefault` and use this to help set
the now negatable value of the `UpdateOnly` flag.

Addresses #775


testdata/script/stack_submit_default_update_only.txt
```diff
@@ -0,0 +1,42 @@
...
+git add feature2.txt
+gs bc feature2 -m 'Add feature 2'
+gs down
+git add feature-update.txt
+git commit --amend
+gs sr
+
+# run stack submit and ensure it only updates the first branch
+gs ss
+
```
klnusbaum, line 34: I'm not sure how to verify that we didn't upload the second branch here.

stack_submit.go
```diff
@@ -13,6 +13,8 @@ import (
 
 type stackSubmitCmd struct {
 	submitOptions
+
+	UpdateOnlyDefault bool `config:"stackSubmit.updateOnly" hidden:"" default:"false"`
```
klnusbaum, line 17: I went with `stackSubmit` instead of `submit` since it seems like this configuration would only be useful for the `stack submit` command. But happy to rename or move this whole thing into the `submitOptions` struct.

testdata/script/stack_submit_default_update_only.txt
```diff
@@ -0,0 +1,42 @@
...
+# create a stack with the first branch submitted
+git add feature.txt
+gs bc feature -m 'Add feature'
+gs bs
+git add feature2.txt
+gs bc feature2 -m 'Add feature 2'
+gs down
+git add feature-update.txt
+git commit --amend
+gs sr
```
haakon-e, line 30: I think this could be shortened to `gs commit amend --no-edit`

testdata/script/stack_submit_default_update_only.txt
haakon-e: I think you could copy this test pretty much verbatim: [`submit_update_only.txt`](https://github.com/abhinav/git-spice/blob/main/testdata/script/submit_update_only.txt). I think the way to test this is by running `gs ll` after `gs ss`, which will show that the second branch does not have a CR number next to the branch name.

abhinav: ^ and logs. I updated to use `gs ls` and to match log output.

stack_submit.go
abhinav: Since this flag should be supported for all multi-branch submit operations, it makes sense to call it submit.updateOnly. (If it was one-command-only, stackSubmit would've made sense.)
I've updated it to do that.

PR #791, opened 2025-08-09
submit: allow default template to be configured

choang:
Add an option to configure the default template file


abhinav (approved):
Nice, thanks!

abhinav (commented):
Actually, seeing as this is _the_ template that will be used if set,
I'm going to rename this to spice.submit.template.
default implies other options.

PR #798, opened 2025-08-10
branch checkout: support remote branches

### LINKED ISSUE #797: Checkout a remote branch with `gs branch checkout`
When I clone a Git repo, it only knows about one branch. I know that I have a feature branch and use `git switch` to checkout it.

I thought that `gs branch checkout` would be a drop-in replacement to checkout the branch and track it too, but:

```
$ gs branch checkout foo
FTL gs: branch "foo" does not exist

$ git switch foo
branch 'foo' set up to track 'origin/itinerary' by rebasing.
Switched to a new branch 'foo'
```

Would it make sense to have the `gs branch checkout` command support this?

abhinav:
'git checkout $branch' falls back to origin/$branch
if it does not exist locally. We should do the same.

Resolves #797


abhinav:

- #798 ◀

PR #815, opened 2025-08-24
test: Fix recurring github/gitlab auth test race

abhinav:
The uitest.RunScripts based tests for testing the authentication forms
have been flaky sine day one because of the concurrency involved.

This replaces that entire mechanism with a more direct approach:
drive the tea.Model directly with a bunch of tea.Msgs.
There's some risk here in that Bubble Tea does not expose
all the underlying messages, but we'll go with this while it works.

Part of the API challenge here was to allow the driver and consumer code
to flow normally (e.g. here's where I specify the events,
and here's where I consume the ui.InteractiveView).
To keep that logic entirely single-threaded, this abuses iter.Seq
to implement a simple coroutine mechanism.

To keep test cases easy to write, we also use autogold
to compare the fixtures instead of managing them manually.


abhinav:

- #815 ◀

PR #829, opened 2025-08-30
feat(gitlab): remove source branch on merge

### LINKED ISSUE #825: GitLab: delete after merge
### Discussed in https://github.com/abhinav/git-spice/discussions/823

<div type='discussions-op-text'>

<sup>Originally posted by **maritz** August 26, 2025</sup>
Currently, it looks like git-spice disables the MR (GitLab CR) setting for a source branch to be deleted after merge.

And it also doesn't automatically delete them when they are locally deleted.

Is this intentional? Am I doing something wrong? This would mean that we accrue *a lot* of stale branches over time, if not manually pruned.</div>

abhinav:
Whether a branch should be deleted after merge
is apparently a MR-level setting, not project-level.
Set it by default for GitLab, and allow opting out
in case someone prefers the old behavior.

Resolves #825

### DIFF
--- a/.changes/unreleased/Changed-20250829-213725.yaml
+++ b/.changes/unreleased/Changed-20250829-213725.yaml
@@ -1,0 +1,7 @@
+kind: Changed
+body: >-
+  gitlab:
+  Submit Merge Requests with `remove_source_branch=true`.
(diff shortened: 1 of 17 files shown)


abhinav:

- #829 ◀

PR #831, opened 2025-08-30
refac(log): extract into handler

abhinav:
Continue the slow migration to the Handler pattern.
Move part of the log command's business logic into a handler.
This better draws the boundary for the upcoming JSON flag.

In porting the code over, I realized that the `isVisible` logic
basically meant that we were querying all branches
and then filtering them in memory at render time.

With the branch graph's new stack querying capabilities,
we're able to avoid querying branches we don't need
when `--all` is not specified.

On a test repository with a 100 branches,
if I run this from a branch that's stacked on top of main,
and has nothing else on top of it (so there won't be a lot to query):

```
❯ hyperfine --warmup 5 'gs ll' '../main/bin/gs ll'
Benchmark 1: gs ll
  Time (mean ± σ):     909.0 ms ±  18.0 ms    [User: 2786.6 ms, System: 3248.9 ms]
  Range (min … max):   896.8 ms … 957.5 ms    10 runs

Benchmark 2: ../main/bin/gs ll
  Time (mean ± σ):     347.5 ms ±   3.5 ms    [User: 951.1 ms, System: 1070.6 ms]
  Range (min … max):   343.7 ms … 351.9 ms    10 runs

Summary
  ../main/bin/gs ll ran
    2.62 ± 0.06 times faster than gs ll
```

(Performance unchanged with --all.)

### DIFF
--- a/.changes/unreleased/Fixed-20250830-143837.yaml
+++ b/.changes/unreleased/Fixed-20250830-143837.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: >-
+  log: When used without `--all`, query only requested branches internally.
+  This was wasted work and slowed the command down in repositories
+  with many branches.
+time: 2025-08-30T14:38:37.576813-07:00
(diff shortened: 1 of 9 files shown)


abhinav:

- #831 ◀

PR #835, opened 2025-08-31
ui: Add commit selection widget

abhinav:
This widget presents the branches and commits in a tree-like structure
similar to `gs ll`, allowing users to select a particular commit.

Example:
commit-widget

PR #834, opened 2025-08-31
git: add analog for 'git write-tree'

abhinav:
Adds a new WriteIndexTree method to git.Worktree
that writes the current index to a new tree object.
This is necessary for implementing low-level operations
on staged changes without committing them.

PR #833, opened 2025-08-31
git: add MergeTree method

abhinav:
This adds a new method `MergeTree` to the `git.Repository` type.
This allows performing merge operations between two branches or trees
without affecting the index or working tree.

In particular, with the --merge-base option (added in Git 2.40),
this allows a differential application of a commit's changes
to another commit or tree.
This feature in particular is useful for implementing
functionality similar to the experimental `git replay` command.

PR #774, opened 2025-07-23
git: extract splitNull function

abhinav:
This was already implemented twice and we were about to do a third.
Extract into a shared scanutil package.

[skip changelog]: no user facing changes

PR #836, opened 2025-08-31
experiment: gs commit fixup

abhinav:
Implements a new 'gs commit fixup' command
that acts like --amend for a commit down the stack.

By default, it presents a nice interactive prompt
for selecting a downstack commit to amend.
You can also specify the commit hash directly,
but that leads to an expensive search across all branches.
I don't have an obvious alternative to that yet.

Up until the final rebase and upstack restack,
the command operates without a working tree.
This requires at least Git 2.45 for this to work.

This is a first step to implementing tree-less operations.

In the future, I expect that restack operations
will be based on the same underlying mechanism.

The new commit will re-use metadata from the original commit.
We don't currently offer a means to alter the commit message.
I think doing that with commit-tree will require
re-creating the 'git commit' editing experience,
which is out of scope for this initial implementation.

The staged changes are squashed into the target commit.
This can be made optional in the future.

This command is gated behind the "commitFixup" experiment.


abhinav:

- #774
    - #833
        - #834
            - #835
                - #836 ◀

PR #841, opened 2025-09-01
all; ioutil.TestLogWriter -> t.Output

abhinav:
Go 1.25 added t.Output(), which produces an io.Writer
that will write to the test log.

This drops the ioutil.TestLogWriter dependency
in favor of t.Output().


abhinav:

- #841 ◀

PR #843, opened 2025-09-01
fix(rebase): flaky TestRebase_deliberateInterrupt

abhinav:
TestRebase_deliberateInterrupt flaked in some environments
because GIT_EDITOR was set, which caused it to use a command
that was not mockedit.


abhinav:

- #843 ◀

PR #859, opened 2025-09-14
branch split: Allow using original name for mid commits

### LINKED ISSUE #411: `gs branch split` prompt which commits (upper or lower) retain the existing branch
Assume I have a stack `master` => `featureOne` => `featureTwo` => `featureThree`
- If I would like to split up `featureTwo` into `featurePreTwo` => `featureTwo`, `gs branch split` works great. The command assumes that the topmost set of commits (furthest away from `master`) will keep the original branch name.
- If I would like to split up `featureTwo` into `featureTwo` => `featurePostTwo`, I'm in trouble. There is no (easy) way to tell `gs branch split` that the bottommost set of commits retain the original branch metadata, and to create a new branch from the topmost set of commits.

I have to resort to naming the bottom branch `temp`, and then use `gs branch` rename to get my desired stack, but then I also have to `gs branch untrack && gs branch track` both branches to bust the existing PR references so things are in a good state.

abhinav:
Today, 'gs branch split' requires branch names for intermediate commits
to be unique and different from the original branch name.
This is good for when the split commit is intended to precede
the original branch:

    o (main)                       o (main)
     \                              \
      A --> B (feature)    =>        A (feature-pre)
                                      \
                                       B (feature)

However, this does not work for cases where the split
should take over the original branch, such as:

    o (main)                       o (main)
     \                              \
      A --> B (feature)    =>        A (feature)
                                      \
                                       B (feature-post)

This commit adds support for reusing the original branch name
for an intermediate commit during a branch split.

This works for interactive and non-interactive modes.

- For interactive mode, if any of the intermediate commits
  request the original branch name,
  a new field is presented to name the original HEAD commit.
- For non-interactive mode, if any of the split points
  reuse the original branch name,
  it is required that a split point for the HEAD commit
  is also provided with a different name.

Resolves #411

### DIFF
--- a/.changes/unreleased/Added-20250914-095715.yaml
+++ b/.changes/unreleased/Added-20250914-095715.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'branch split: Allow reusing original branch name for intermediate commits during split'
+time: 2025-09-14T09:57:15.397904-07:00
--- a/branch_split.go
+++ b/branch_split.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	"go.abhg.dev/gs/internal/git"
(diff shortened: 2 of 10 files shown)


abhinav:

- #859 ◀

PR #861, opened 2025-09-15
refac(log): Extract presentation logic into logPresenter

abhinav:
Extracts the logic to render a log of branches
into a separate interface so that we can implement
the upcoming JSON output format using the same underlying data.

[skip changelog]: no user facing changes.


abhinav:

- #861 ◀

PR #866, opened 2025-09-18
feat: added status symbol to log command (--status)

svapnil:
This pull request adds a colored "status" symbol to the terminal.

We're exposing the Status enum in gs log, both in the visualization and in the JSON

Here is a (redacted) version of what this looks like in practice:

image

Eventually, it would be helpful if we added additional states like (CI_IN_PROGRESS (yellow), CI_FAILED(red)) to Status we can use "gs ls" to get a complete picture on how our stacks are doing.


abhinav (commented):
Just looked over this. Nice!
I'm supportive of this overall.
I don't love that we'll make a network call per `gs ls` with this, so it definitely makes sense for it to be opt-in. 
Give me a little time to poke at this and get it merge-ready.

Just to share some reasoning:
I'm imagining most users will not enable this in their configuration, but ask `gs ls` to fetch this information on demand, so I might also add a short flag to enable this.
In lieu of "format" for change status information, I might just make it on or off for now: off = no fetch, on = fetch and show symbols. In the future, if there's need for customizing the symbols (e.g. if someone prefers words), there may be a separate configuration option for the format used when this information is requested.

abhinav:
Sure, go ahead make the bool change. 

I'll poke around with this later today and might push a few more changes.
(I want to make sure the UX feels right to me.) I can't promise I'll merge it this weekend as I've got a bunch of stuff going on.

I'll probably also create a couple test scripts to verify the functionality. Feel free to add them if you'd like to try but not required. (I'm happy to add them myself.)

abhinav:
Thanks, @svapnil!

PR #868, opened 2025-09-20
fix(commit fixup): handle autostashing dirty changes

abhinav:
If there are unstaged changes in the worktree when using commit fixup
to fix up a downstack branch, we should not pop the autostashed changes
until after the upstack restack is complete,
and we've switched back to the original branch (not the fixed branch).

Ref: https://github.com/abhinav/git-spice/discussions/867#discussioncomment-14465048

[skip changelog]: New entry not required as fixup hasn't been released yet.


abhinav:

- #868 ◀

PR #873, opened 2025-09-25
doc: Add a tapes/ folder

abhinav:
We can put VHS tapes for demos here.
These will not be rendered at doc build time
as that'll change literally every build (git hashes).

We can set up infrastructure to render them deterministically
(similarly to test scripts) in the future if we really want to,
but even then because the VHS tapes run in real time,
that will significantly slow down docs (and will likely be flaky in CI).

So the tapes folder is just for future reference
to recreate any demos if we want to.


abhinav:

- #873 ◀

PR #879, opened 2025-09-28
refac(track/Handler): AddBranch => TrackBranch

abhinav:
Renames AddBranch to TrackBranch to better reflect its purpose.

PR #886, opened 2025-10-06
feat: add 'downstack track' command

### LINKED ISSUE #300: branch track: track a whole untracked stack
Scenario:

```
git checkout main
git checkout -b feat1
git checkout -b feat2
```

At that point, we'd like `gs branch track` or an equivalent command to track `main -> feat1 -> feat2`, not `main -> feat2`.
This would have to be opt-in to handle the case where the intermediate branch is untracked on purpose.

Possibilities:

- `gs downstack track`
- `gs branch track --all`

abhinav:
Add a new 'gs downstack track' command.
This allows users to track a manually created stack of branches
in one go by traversing the commit graph downwards.
It's possible to skip branches in the process,
e.g. if some intermediate branches aren't intended to be tracked.

Resolves #300


abhinav:

- #879
    - #880
        - #886 ◀

PR #901, opened 2025-10-18
Add prompt for stack/upstack restack from trunk

### LINKED ISSUE #898: gs {stack, upstack} restack on trunk should verify
It's rare to intentionally want to run `gs stack restack` from the trunk branch
as that restacks everything, and will often result in conflicts that the user doesn't want to deal with.
Keeping in line with `gs commit amend` on trunk and other similar prompts,
we should prompt the user to verify that they mean to run the operation on main.

Note that for users who do mean it, and do it regularly, and don't want to be prompted,
we do have the `gs repo restack` operation which will restack all tracked branches.

abhinav:
When running 'stack restack' or 'upstack restack' from the trunk branch,
prompt users to confirm the operation in interactive mode,
and log a warning in non-interactive mode.

This helps prevent accidental restacking of all managed branches.

Users who actually meant to do this and don't want to see the prompt
can use the 'gs repo restack' command instead,

Resolves #898

### DIFF
--- a/.changes/unreleased/Added-20251018-115634.yaml
+++ b/.changes/unreleased/Added-20251018-115634.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: >-
+  {stack, upstack} restack: prompt for confirmation when invoked from trunk
+  to prevent accidental full-repository restacking.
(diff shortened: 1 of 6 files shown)


abhinav:

- #901 ◀

PR #902, opened 2025-10-18
fix(mise): don't pass verbose flag to 'go test'

abhinav:
go test output should not be suppressed even in non-verbose mode,
because in that case, gotestsum doesn't see anything
(and won't report any tests as run even if they did).

`go test` should always be verbose,
and the `--format` flag to gotestsum
controls how much output is shown.


abhinav:

- #902 ◀

PR #919, opened 2025-10-29
fix(repo sync): Don't delete branch if open PR is found

benrucker:
## Proposed change

If you're re-using a branch name, this PR reduces the likelihood that `git-spice` prompts you to delete it.

This helps reduce friction for new users in old/large codebases, where they're probably using some combination of `git-spice` and regular `git`+web UI, by making it less likely that they'll accidentally delete one of their branches.

It does this by changing how `gs repo sync` handles _externally_ created changes such that if it finds an open change, it will not prompt the user to delete the local branch. So if you're re-using a branch name for a change that was merged in the past _and_ you already have a change open for the new branch, then `git-spice` won't mark the branch as merged or prompt you to delete it. However, this PR does not help if you haven't externally opened a change request or you haven't run `gs branch submit`.

This functionality is achieved by making the following changes:

- When checking for externally created changes:
  - Instead of loading only `merged` changes, load all types of changes
  - If any of the 3 loaded changes are open, set the open one as the change for the branch
  - If no open changes are found and a merged change is found, set that one as the change for the branch

### Test

I've made a change to the existing test file for how `repo sync` handles external branches to test this behavior. I am happy to instead make a new test file if desired!

### Caveat

I'm not proficient in Go, so the initial version of the code and the test was written with Claude Sonnet 4.5. I reviewed and refactored the code to the best of my ability, but I am willing to iterate on this or change direction.

### Questions

The original code loaded 3 changes, and I'm assuming they're loaded in reverse chronological order (newest first). Is this true?

I think only loading 3 is fine, although ideally we could filter to PRs that are either open or merged, but not closed. I'm thinking that it's unlikely that a user has 3 closed PRs for the same branch that are newer than the open one they care about.

### DIFF
--- a/.changes/unreleased/Fixed-20251028-143010.yaml
+++ b/.changes/unreleased/Fixed-20251028-143010.yaml
@@ -1,0 +1,5 @@
+kind: Fixed
+body: >-
+  Prevent `gs repo sync` from prompting to delete a branch
+  when both a merged and an open PR are present for it.
+time: 2025-10-28T14:30:10-07:00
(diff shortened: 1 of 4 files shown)


abhinav (approved):
Awesome, thanks!

Issue #934, opened 2025-11-11
`gs bs -am` requires `--commit` before it uses the message to automatically create a branch name


choang:
I like having the following create a branch for me:
```
gs bs -am "my stuff"
```

instead, I get this error:
```
FTL gs: a branch name is required with --no-commit
```

and I have to add the `--commit` for it to work
```
gs bs --commit -am "my stuff"
```

PR #935, opened 2025-11-11
branch create: generate branch name when -m implies --commit

abhinav:
Fixes https://github.com/abhinav/git-spice/issues/934

When branchCreate.commit is configured to false (making --no-commit
the default), using -m flag now properly enables commits and allows
automatic branch name generation from the commit message.

The fix moves the logic that sets cmd.Commit = true when -m is
provided before the validation check that requires a branch name
with --no-commit. This ensures the -m flag overrides the
configuration as intended.

**AI assistance**:
The test, the fix, and the changelog entry were entirely composed by an LLM.
A human operator verified them afterwards, editing the test and the changelog slightly.

### DIFF
--- a/testdata/script/issue934_bc_am_with_no_commit_default.txt
+++ b/testdata/script/issue934_bc_am_with_no_commit_default.txt
@@ -14,14 +14,7 @@ gs repo init
 # Configure branchCreate.commit to false (--no-commit as default).
 git config --local spice.branchCreate.commit false
 
-# Create a file with changes to commit.
-cp $WORK/extra/feature.txt feature.txt
 git add feature.txt
-
-# Run 'gs bc -am "Add feature"' which should:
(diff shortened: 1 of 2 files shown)


testdata/script/issue934_bc_am_with_no_commit_default.txt
```diff
@@ -0,0 +1,47 @@
...
+# Verify that the commit was made with the correct message.
+git log --format='%s' -n 1
+cmp stdout $WORK/golden/commit-message.txt
+
+# Verify the file is committed and working tree is clean.
+git status --porcelain
+! stdout .
+
+-- repo/.git/.keep --
+-- extra/feature.txt --
```
abhinav:
```suggestion
-- repo/feature.txt --
```

PR #941, opened 2025-11-15
fix(submit): Prompt for draft status on all branches in stack

### LINKED ISSUE #936: gs stack submit only asks for draft once
When submitting multiple CRs from a stack, it only asks for whether the CR should be in draft mode for the first one.

Example output with 3 branches of which the second and third were not submitted previously:

```
/home/me/projectDir # gs ss
INF CR !70 is up-to-date: https://gitlab.com/mycompany/mygroup/myproject/-/merge_requests/70
Title: feat: add cool thing
Title: feat: add cool thing 
Template: Default
Body: Press [e] to open nano or [enter/tab] to skip
Title: feat: cool thing 
Template: Default
Body: Press [e] to open nano or [enter/tab] to skip
Draft: [y/N]
INF Created !72: https://gitlab.com/mycompany/mygroup/myproject/-/merge_requests/72
Title: chore: bump version to 1.2.3 
WRN Template "merge_request_templates/empty.md" not foundTitle: chore: bump version to 3.3.0 
Template: Default
Body: Press [e] to open nano or [enter/tab] to skip
Title: chore: bump version to 1.2.3 
Template: Default
Body: Press [e] to open nano or [enter/tab] to skip
INF Created !73: https://gitlab.com/mycompany/mygroup/myproject/-/merge_requests/73
```

This was actually nice in this case, because I would've pressed N anyway, but seems like it is unintended behavior.

abhinav:
When submitting multiple branches in a stack interactively,
the draft status prompt was only shown for the first branch.
Subsequent branches would skip the prompt entirely.

The issue occurred because `SubmitBatch` reused the same
`Options` struct across all branches in the loop.
When processing the first branch,
`submitBranch` would set `opts.Draft` from `nil` to a pointer.
For subsequent branches,
the `opts.Draft != nil` check would fail,
preventing the draft prompt from appearing.

Fix this by creating a shallow copy of the `Options` struct
for each branch before passing it to `submitBranch`.
This ensures that modifications to the options
only affect the current branch being processed.

Resolves #936

### DIFF
--- a/.changes/unreleased/Fixed-20251114-191245.yaml
+++ b/.changes/unreleased/Fixed-20251114-191245.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'submit: Fix draft prompt being skipped for branches after the first one when submitting a stack interactively'
+time: 2025-11-14T19:12:45.004898-08:00
--- a/internal/handler/submit/handler.go
+++ b/internal/handler/submit/handler.go
@@ -292,10 +292,13 @@ func (h *Handler) SubmitBatch(ctx context.Context, req *BatchRequest) error {
 
 	var branchesToComment []string
 	for _, branch := range req.Branches {
+		// Shallow copy the options because submitBranch may modify them.
+		opts := *opts
+
 		status, err := h.submitBranch(
 			ctx,
 			branch,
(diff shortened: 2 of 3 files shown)


abhinav:

- #941 ◀

PR #953, opened 2025-11-30
doc: fix removeSourceBranch config version

abhinav:
`gs:unreleased` is wrong so it was never updated.


abhinav:

- #953 ◀

PR #960, opened 2025-12-03
test: Fix flaky stack_submit_with_labels test on Windows

abhinav:
Replace exact stderr comparison with pattern matching
in the `stack_submit_with_labels.txt` test script.
The test was intermittently failing on Windows
due to warning messages about template caching
that appear non-deterministically,
similar to #956 and #957.

Instead of comparing the entire stderr output
against a golden file,
the test now checks for the presence
of the three "Created #X" messages,
which is sufficient to verify the command succeeded.

Co-Authored-By: Claude <noreply@anthropic.com>


abhinav:

- #960 ◀

PR #961, opened 2025-12-04
test(commit split): fix for Git 2.52

abhinav:
Prompt for 'git add -p' changed in Git 2.52, so tests are now failing.
Update the fixtures and skip these tests on older Git versions.


abhinav:

- #961 ◀

PR #983, opened 2026-01-04
github: Use REST for requesting reviews

### LINKED ISSUE #972: Document required permissions for adding team as reviewer
```
add reviewers to PR: resolve reviewer IDs: lookup team org/"org/engineering": query team: Post "https://api.github.com/graphql": organization.team: FORBIDDEN: Resource not accessible by integration
```

I'm using OAuth App integration

abhinav:
## Why use GitHub v3

Using the GraphQL API to request reviews for a PR
requires resolving user names and team names to GraphQL node IDs first.
This requires the `read:org` permission scope,
which is not something git-spice has by default.
(The OAuth and GitHub Apps do not request that scope.)

Rather than expanding the permission scope
(which would require users to re-authorize git-spice),
switch to using the GitHub REST API for requesting reviews.
As this is just a PR-level operation, per the documentation,
this only requires the `write:pull_request` scope.
(It is quite annoying that the GraphQL API provides
no other way to do this, and there are operations
that are not possible to do with REST alone,
so you have to mix and match both APIs.)

## Why hand-write a client

This change does not use the go-github library directly,
and instead implements a hand-rolled minimal GitHub REST API client.
The reason for this is this design choice that go-github makes:

    type Client struct {
      Issues              *IssuesService
      PullRequests        *PullRequestsService
      RepositoriesService *RepositoriesService
      ...
    }

Where each `*Service` type has many many methods.

Normally, the Go linker would be able to strip out
unused methods and types from the final binary.
But because we use Kong for CLI parsing,
we have reflect `MethodByName` in our call graph,
and when that's the case the linker cannot say with certainty
which methods are used and which are not,
and we end up pulling in the entire go-github library,
which inflates the binary size by several megabytes.

(The GitLab client library also suffers from this,
and we'll probably replace it with a hand-rolled minimal client
in the future as well.)

## AI Assistance

An LLM was used to put together the initial version of the client,
which was then refined by a human developer.
The unit tests for the new functionality were authored by an LLM
with very specific instructions provided by a human developer.
No significant changes were made to these after generation.

## Manual verification

Draft until the following has been performed.

- [x] With a development build of this change,
      actually verify that requesting reviews from teams works
      without `read:org` scope.

Resolves #972


abhinav:

- #983 ◀

PR #993, opened 2026-01-15
fix(xec): EditCommand should forward stdout/stderr/stdin

abhinav:
EditCommand should not leave stdout/stdin/stderr unset
as these will make the editor not work.

[skip changelog]: bug wasn't released yet


abhinav:

- #993 ◀

PR #1011, opened 2026-02-03
Fix Flaky Submit Test

ed-irl:
This fixes a flaky test assertion that relies on strict string comparison. The root cause is a warning that prints during template listing.


abhinav (approved):

abhinav:
Thanks!

PR #1013, opened 2026-02-06
forge: extract shared URL parsing utilities

ed-irl:
Extract duplicate URL parsing code from github and gitlab forges into a new internal/forge/urls package.

The new package provides:
- HasGitProtocol: check if URL has a known git protocol
- Parse: parse git remote URLs, normalizing SCP-style SSH syntax
- StripDefaultPort: remove default ports when base URL doesn't specify one
- MatchesHost: check if hosts match including subdomains
- ExtractPath: extract owner/repo from URL path

This reduces code duplication and will allow the bitbucket forge to reuse the same URL parsing logic.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -329,6 +329,9 @@ Where:
     or command domains (e.g., "submit", "github").
   - See CHANGELOG.md for existing patterns.
 
+To skip the changelog check for internal changes, refactors,
+or test-only changes, include `[skip changelog]: <cause description>` in the PR description as a trailer.
+
 # Code Quality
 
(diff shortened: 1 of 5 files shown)


CLAUDE.md
```diff
@@ -343,6 +346,23 @@ Where:
...
+- **Package naming**
+
+  Do not pluralize package names.
+  Use singular nouns or compound names instead.
+
+  ```
+  // BAD: pluralized package names
+  urls
+  utils
+  helpers
```
abhinav: ❤️ 

thanks!

PR #1015, opened 2026-02-07
forge: add comment modularity interfaces

ed-irl:
Adds support for creating direct CR links instead of relying on forge's auto-linking when they encounter a `#nnn` magic string. Prerequisite for implementing bitbucket forge (which does not auto link magic strings).

[skip changelog]: Internal-only refactor to allow forges that don't auto-link CR's when they see a `#nnn` magic string


abhinav (approved):
Very clean, thanks!
I appreciate that the reasoning for adding these is clear.

PR #1016, opened 2026-02-08
internal: add stack-based signal handler

abhinav:
Add a Stack type that manages per-signal handler stacks.
Handlers are invoked in LIFO order,
and signal registration is managed lazily:
Notify on first push, Stop+Reset when the stack empties.

This allows multiple components to register handlers
for the same signal without clobbering each other.


abhinav:

- #1016 ◀

PR #1027, opened 2026-02-15
branch checkout: Add `trackUntracked` config option

### LINKED ISSUE #930: `gs branch checkout`: Ability to track untracked branches by default
I have the opposite use case of what https://abhinav.github.io/git-spice/cli/config/#spicebranchcheckouttrackuntrackedprompt solves: I frequently checkout untracked branches that I _want_ to track, but I find the prompt repetitive.

The `spice.branchCheckout.trackUntrackedPrompt` option also confused me because I thought setting it to `false` would suppress the prompt while keeping the behaviour that the prompt defaulted to (which is to track the untracked branch). Instead, the prompt is suppressed and the branch is left untracked.

abhinav:
Replace the boolean `spice.branchCheckout.trackUntrackedPrompt`
with a three-mode `spice.branchCheckout.trackUntracked` option
that accepts `prompt` (default), `never`/`false`, or `always`/`true`.

This gives users who frequently checkout untracked branches
the ability to auto-track without being prompted every time,
addressing the gap where the old boolean only supported
"prompt" or "don't track" modes.

The deprecated option is still honored when set:
`true` maps to `prompt`, `false` maps to `never`,
and a deprecation warning is logged.
Deprecated config entries are also hidden
from help and CLI reference output.

Resolves #930

### DIFF
--- a/.changes/unreleased/Changed-20260214-080054.yaml
+++ b/.changes/unreleased/Changed-20260214-080054.yaml
@@ -1,0 +1,7 @@
+kind: Changed
+body: >-
+  branch checkout: Add 'spice.branchCheckout.trackUntracked' configuration option to control tracking of untracked branches.
+  Defaults to prompting the user,
+  but can be set to 'always' to always track untracked branches,
+  or 'never' to not track nor prompt.
+time: 2026-02-14T08:00:54.574347-08:00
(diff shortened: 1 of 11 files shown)


abhinav:

- #1027 ◀

PR #1046, opened 2026-03-02
feat: add 'branch diff' command

### LINKED ISSUE #1040: show changes in a branch
I'm loving git-spice and slowly learning more and more of its tricks.

I would like a way to easily view the changes proposed by the current branch please. Maybe such thing already exists and I did not see it in the help.

Consider: `git-spice branch diff` should show the changes from HEAD compared to the base branch.

`--branch foo` could be used to compare an arbitrary branch `foo` to `foo`'s own base, similar to how `--branch` works in other `git-spice` commands like `onto`.

Maybe allow opportunity to compare arbitrary branches, but that may not be helpful compared to standard git.

Probably have to account for situations where downstack restack is needed. Maybe fail in such case and/or prompt user to restack?

Thanks!

abatilo:
## Summary

- Add `gs branch diff` (alias `gs b di`) that shows the diff between a tracked branch and its base branch in the stack using `git diff base...branch` (triple-dot syntax)
- Defaults to the current branch; use `--branch` to target a different one
- Errors clearly on untracked branches

Resolves #1040

## Test plan

- [x] `testdata/script/branch_diff_basic.txt` — basic diff against base
- [x] `testdata/script/branch_diff_flag.txt` — `--branch` flag targeting a different branch
- [x] `testdata/script/branch_diff_untracked.txt` — error on untracked branch
- [x] `mise run build` passes
- [x] `mise run lint` passes

### DIFF
--- a/doc/includes/cli-reference.md
+++ b/doc/includes/cli-reference.md
@@ -951,6 +951,25 @@ Use 'gs upstack onto' to also move the upstack branches.
 
 **Configuration**: [spice.branchPrompt.sort](/cli/config.md#spicebranchpromptsort)
 
+### git-spice branch diff {#gs-branch-diff}
+
+```
+gs branch (b) diff (di) [flags]
+```
+
+Show diff between a branch and its base
+
+Shows the diff between a branch
+and its base branch in the stack.
+This is equivalent to running
+'git diff base...branch'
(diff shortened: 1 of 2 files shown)


abhinav (approved):
Thanks!

PR #1054, opened 2026-03-10
ui: migrate charmbracelet stack to v2

abhinav:
Upgrades to charmbracelet's v2 libraries
(lipgloss, bubbletea, and bubbles, https://charm.land/blog/v2/).

This required significantly reworking our rendering pipeline,
as well as some workarounds on the test/simulation side
to accommodate improvements made on the Bubble Tea side.

## Render-time styling

Upgrade the TUI stack to `bubbles`, `bubbletea`, and `lipgloss`
v2,
and reshape the rendering pipeline around the API changes those
releases require.

Lip Gloss v2 resolves adaptive colors against the active output stream
at render time,
so the UI now carries semantic `ui.Style` and `ui.Theme` values until
it writes output.
That keeps call sites expressive while letting the output layer decide
how much styling the destination can support.

## Bubble Tea v2 integration

Bubble Tea v2 also changes the terminal contract.
Forms now implement `View() tea.View`,
key handling moves to `tea.KeyPressMsg`,
and startup sizing is configured through program options.
Those updates are mostly mechanical,
but they require the interactive layer to pass theme and window-size
state through every field consistently.

## Test harness and terminal emulation

The migration also forced the scripted UI harnesses to behave more like
real terminals.
The emulator now preserves raw-terminal newline semantics,
treats tabs as cursor movement,
and waits for the screen to settle before assertions snapshot the
result.
That keeps Bubble Tea v2 redraw bursts from leaking intermediate frames
into fixtures.

Robot-driven prompt fixtures need the same output-capability handling.
`lipgloss.Style.Render` still returns ANSI sequences even when tests
are recording prompt text into plain files.
`uitest.RobotView` now captures prompt and log output through a
`colorprofile.Writer` forced to `NoTTY`,
so script fixtures observe the same unstyled text that a non-terminal
destination would receive.
A regression test locks that behavior down for future UI changes.

The PTY-based `with-term` harness also needs a deterministic theme.
Its scripted terminal does not answer background-color probes,
so detecting the active theme can block interactive startup before the
first prompt is drawn.
`ui.detectTheme` now honors a test-only `__GIT_SPICE_THEME` override,
and `with-term` sets it explicitly so PTY-driven script tests can boot
without stalling on terminal queries.

## Output capability detection

The migration also changes how non-interactive output is written.
`ui.NewFileView` wraps outputs with `colorprofile.Writer`
and records the detected theme alongside the writer.
That preserves colored branch-tree and log output when the destination
supports it,
while still stripping or down-sampling styles when it does not.

### DIFF
--- a/internal/ui/form.go
+++ b/internal/ui/form.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
 	"charm.land/bubbles/v2/key"
@@ -163,6 +164,10 @@ type FormRunOptions struct {
 	// If both are greater than zero, these are applied with tea.WithWindowSize.
 	Width, Height int
 
+	// TERM overrides the terminal type that Bubble Tea uses
+	// for capability detection.
+	TERM string
+
 	// WithoutSignals requests that the form not register signal handlers.
 	WithoutSignals bool
 }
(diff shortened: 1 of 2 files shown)


abhinav:

- #1054 ◀

PR #1060, opened 2026-03-12
test/with-term: log output before close

abhinav:
We're seeing flakiness on tests relying on with-term.
Output doesn't include what went wrong because it's all going
to the terminal emulator.

Log the output on failure.


abhinav:

- #1060 ◀

PR #1067, opened 2026-03-16
Add comment counts option to git-spice log command

ed-irl:
Add an optional comment count output to the log command, to allow users to see the total number of comments and number resolved on a particular branch. Because this command uses network to fetch comment counts (unlike the "normal" log command) it is under a flag.


abhinav:
FYI, @ed-irl I've moved this to #1087 while I'm iterating on it because this PR does not have Maintainers Can Modify enabled. I'm flipping this PR to draft to avoid accidentally merging it instead.

PR #1073, opened 2026-03-25
sync: retarget upstack PRs after deleting merged branches

ed-irl:
Fixes a "bug" I discovered while testing my branch merge implementation where syncs could cause weird state that would mess up future merges.

- After `repo sync` deletes merged branches, surviving upstack PRs are retargeted on the forge so their base matches the updated local state
- Walks the base chain to find the nearest surviving ancestor, falling back to trunk for cycles or missing entries
- Covers multi-level deletions (e.g., A and B merged, C retargets to main)

### DIFF
--- a/internal/handler/sync/handler.go
+++ b/internal/handler/sync/handler.go
@@ -355,6 +355,15 @@ func (h *Handler) SyncTrunk(ctx context.Context, opts *TrunkOptions) error {
 		return err
 	}
 
+	// Retarget surviving upstack PRs on the forge
+	// so their base matches the updated local state.
+	if h.RemoteRepository != nil {
+		h.retargetUpstackChanges(ctx,
+			collectRetargetCandidates(
+				branchesToDelete, candidates, trunk,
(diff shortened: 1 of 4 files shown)


abhinav:
Thanks, @ed-irl!
So just to clarify/confirm the behavior:
git-spice already retargets upstack branches when their bases are deleted, but it does so entirely locally.
The focus of this change is to specifically also update the forge?
The next `submit` for those branches would also update the base branch in the forge.
I would appreciate some more context on what issue we're running into here.
(Context on my hesitance: repo sync is currently read-only for the forge. This would add a write, so I want to make sure we have a good reason to.)

ed-irl:
Closing in favor of #1150

abhinav:
Thanks, the explanation in https://github.com/abhinav/git-spice/pull/1150 is helpful. 
I think you forgot to press close, so I'll close in favor of https://github.com/abhinav/git-spice/pull/1150.

PR #1074, opened 2026-03-25
git: retry on index.lock contention, set GIT_OPTIONAL_LOCKS=0

ed-irl:
Fix for lockfile issues I encountered while developing a VSCode extension for git-spice. The extension makes background reads and they occasionally cause git-spice to fail. This stops advisory index lock creation for commands that don't need it, and adds backoff incase another process has created a lock.

- Read-only git commands now set GIT_OPTIONAL_LOCKS=0 to avoid contending with concurrent index lock holders
- Write commands (checkout, commit, reset, etc.) automatically retry with exponential backoff when they fail due to index.lock contention
- Rebase operations have specialized recovery that detects whether the sequencer started and either continues or re-runs the rebase after the lock clears
- New spice.indexLockTimeout config option controls retry timeout (in milliseconds; default 5s, 0 disables, negative retries indefinitely)
- Generic retry infrastructure added to xec.Cmd via RetryPolicy with match function, timeout, and exponential backoff

[skip changelog]: fix only

### DIFF
--- a/internal/git/cmd.go
+++ b/internal/git/cmd.go
@@ -6,7 +6,10 @@ package git
 
 import (
 	"context"
+	"errors"
 	"strings"
+	"sync"
+	"time"
 
 	"go.abhg.dev/gs/internal/silog"
 	"go.abhg.dev/gs/internal/xec"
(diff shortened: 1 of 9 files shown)


abhinav:
Just a quick update: haven't forgotten about this. This has been a busy week.
 
This is an overall good direction. I'm making some changes but this should help with the issues that have been reported with concurrent operations in the past.

abhinav:
Merged in https://github.com/abhinav/git-spice/pull/1106

PR #1098, opened 2026-04-06
test(ui): drop midterm TERM workaround

abhinav:
Commit 7ba2fc79 migrated to Bubble Tea v2,
but it had to force `TERM=xterm-256color`
in emulator-backed tests because with `TERM` unset,
Bubble Tea v2 could take an insert-mode redraw path.

In insert mode, the terminal is expected to shift the active row right
instead of overwriting it. `midterm` did not support this at the time,
so the `TERM` workaround was required.

Insert mode support was added to midterm in vito/midterm#2,
and recently merged, so we can switch to it and drop the workaround.


abhinav:

- #1098 ◀

PR #1099, opened 2026-04-11
internal/git: reintroduce gitCmd

abhinav:
For #1074, we need to add a bunch of functionality
around command execution for Git operations.
Introduce the gitCmd wrapper around xec.Cmd to make these easier.

[skip changelog]: no user facing changes

PR #1100, opened 2026-04-11
refactor(git): pass subcommands explicitly

abhinav:
Make `newGitCmd`, `Repository.gitCmd`, and `Worktree.gitCmd`
accept the Git subcommand separately from the remaining argv.

This removes the need to try to extract the subcommand
from the leading arguments, which was hacky and fragile.
This obviates the need for gitCmd.WithLogPrefix,
since the log prefix is now always "git <subcmd>".

[skip changelog]: no user visible change


abhinav:

- #1099
    - #1100 ◀
        - #1101

PR #1101, opened 2026-04-11
git: set GIT_OPTIONAL_LOCKS=0 for read-only commands

abhinav:
(Extracted from #1074)

Read-only git subcommands now receive GIT_OPTIONAL_LOCKS=0
to avoid contending with concurrent writers.

Per Git documentation,

    If this Boolean environment variable is set to false,
    Git will complete any requested operation
    without performing any optional sub-operations
    that require taking a lock.
    For example, this will prevent git status
    from refreshing the index as a side effect.
    This is useful for processes running in the background
    which do not want to cause lock contention
    with other operations on the repository.
    Defaults to 1.


abhinav:

- #1099
    - #1100
        - #1101 ◀

PR #1106, opened 2026-04-15
git: retry on index.lock contention (#1074)

abhinav:
Adapted from #1074

Write subcommands that fail due to index.lock contention
are automatically retried with exponential backoff,
controlled by spice.git.indexLockTimeout (default 5s).


abhinav:

- #1099
    - #1100
        - #1101
            - #1106 ◀

PR #1113, opened 2026-04-20
fix(ui): don't eat letter from next shell command

### LINKED ISSUE #1079: `gs repo restack` eats first letter of subsequent command
There is a sizeable chance this is caused by my environment, but I've been running into this problem for some time, and it seems to only happen with git-spice.

I use fish shell and have an abbreviation `gsrr` that expands to `gs repo restack` (although I don't see how the abbreviation is relevant to the situation -- it should be just like running `gs repo restack` directly, unless there is a bug in fish).

While this runs, I will often start typing my next command. However, git-spice consistently eats the first character of the next command. This is hard to visualize, but for example:

Image

I start typing `gco dev` as the command is running. Notice that only `co ...` appears. I think git-spice is somehow consuming the `g`. When `gs repo restack` completes, what I typed is in the command line already, but the first `g` is missing.

abhinav:
**Context**

In #1079, a user reported a weird issue:
if they run `gs repo restack` and then immediately type the next command
(while restack is still running),
the first letter of that next command can get lost.

For example:

    > gs repo restack <ENTER>
    gpushf --all <ENTER>
    [.. restack output..]
    Restacked N branches
    > pushf --all
    fish: Unknown command: pushf

While the issue was reported for fish,
I was able to reproduce it in bash and zsh as well.

The included test reproduces the issue as described in a real PTY,
but against bash (as that's available by default on Linux/macOS runners).
The test roughly operates like so:

- Set up a repository with a tracked stack of branches
- Advance trunk so the stack needs to be restacked
- Start an interactive bash session attached to a PTY
- Run `gs repo restack` in the bash session
- Wait a short time (the issue reproduced reliably at 50ms for me)
- Inject the first letter of the next command (e.g. "e" for "echo")
- Wait for restack to finish and the prompt to return
- Type the rest of the next command (e.g. "cho $VAR\n")

The output SHOULD contain the value of "$VAR",
but without the fix bash attempts to run "cho $VAR"
and reports "command not found".

**Issue**

The root cause of the issue is UI theme detection.
We probe the terminal to pick between light/dark themes,
and that probe involves temporarily switching the TTY into raw mode.

**Fix**

Stop probing the terminal eagerly during startup.
Instead, compute the theme lazily, the first time it is needed,
which only happens when rendering colored output or prompting for input.

There's still a non-interactive code path that will query the theme,
but it won't trigger until it is rendering output.
(This reduces the race window, but it doesn't eliminate it entirely.)

Resolves #1079

### DIFF
--- a/.changes/unreleased/Fixed-20260419-162108.yaml
+++ b/.changes/unreleased/Fixed-20260419-162108.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: >-
+  ui: Fix interactive shells occasionally dropping the first character
+  of the next command when typing begins before a git-spice command has
+  fully returned control to the shell.
+time: 2026-04-19T16:21:08.637012-07:00
--- a/internal/ui/view.go
+++ b/internal/ui/view.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"io"
 	"os"
+	"sync"
 
 	"github.com/charmbracelet/colorprofile"
 )
@@ -79,7 +80,7 @@ type TerminalView struct {
 	r io.Reader
 	w io.Writer
 
-	theme Theme
+	theme func() Theme
 }
 
 var _ InteractiveView = (*TerminalView)(nil)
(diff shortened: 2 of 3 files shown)


abhinav:

- #1112
    - #1113 ◀

Issue #1120, opened 2026-04-22
auth: Login broken for forges that support GCM


abhinav:
## Reproduction

To reproduce, given a forge that supports GCM:

1. Log out of git-spice
2. Make sure a credential helper is not installed for that forge    
    ```
    ❯ git config --get credential."https://github.com".helper
    [no output]
    ```

3. Run `gs auth login`

This prompts for a username and password instead of presenting login options.

```
❯ gs auth login
Select a Forge: github
Username for 'https://github.com':
Password for 'https://github.com':
```

## Problem

We currently query `git-credential-manager` before anything else for GCM-enabled forges.
This runs the command:

```
echo $'protocol=https\nhost=github.com\n\n' | git credential fill
```

Per git documentation,

```
If the action is fill, git-credential will attempt to add "username" and "password" attributes [..], or by prompting the user.
```

By default, GCM will prompt if a credential helper is not configured for that service.

## Fix

1. We should reconsider unconditionally querying GCM.
    Perhaps we query GCM only if you configure GCM at login.
    This will require storing "use GCM" as the token similarly to "use CLI".

2. Git can be configured to not prompt by setting GIT_TERMINAL_PROMPT=0:

    ```
    ❯ echo $'protocol=https\nhost=github.com\n\n' | GIT_TERMINAL_PROMPT=0 git credential fill
    fatal: could not read Username for 'https://github.com': terminal prompts disabled
    ```

    We should do this regardless of the fix as it will prevent other such issues in the future.

PR #1121, opened 2026-04-22
auth: disable git terminal prompt when loading GCM credentials

abhinav:
Prevent `git credential fill` from prompting for input when no credential helper is configured,
which could hang or block flows that expect non-interactive behavior.

This is easy enough by setting `GIT_TERMINAL_PROMPT=0` when invoking `git credential fill`.

The regression test is pretty narrow, just testing that the env var was set.
Worth considering an end-to-end test for the broken flow as well
as part of the fuller fix for the issue.

Refs #1120

### DIFF
--- a/.changes/unreleased/Fixed-20260422-170432.yaml
+++ b/.changes/unreleased/Fixed-20260422-170432.yaml
@@ -1,3 +1,5 @@
 kind: Fixed
-body: 'auth: Prevent git credential fill from prompting when no credential helper is configured'
+body: >-
+  auth: Disable prompting when querying Git Credential Manager (GCM).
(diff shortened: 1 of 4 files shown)


.changes/unreleased/Fixed-20260422-170432.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Fixed
+body: 'auth: Prevent git credential fill from prompting when no credential helper is configured'
```
abhinav:
```suggestion
body: >-
  auth: Disable prompting when querying Git Credential Manager (GCM).
  With prompting enabled, GCM may block the authentication flow.
```

PR #1122, opened 2026-04-23
auth: stop implicit GCM credential fallback

### LINKED ISSUE #1120: auth: Login broken for forges that support GCM

abhinav:
`gs auth login` could hit `git credential fill` before the user
selected Git Credential Manager.
When no helper was configured,
Git prompted for credentials and bypassed the intended auth flow.

Store GCM as an explicit authentication mode instead of an
ambient fallback,
and load credentials on demand only for tokens that were saved
with that mode.
Update the auth docs to match that contract,
and move the regression tests to package-local `TestMain` fake
`git` helpers so they exercise the same path without shell shims.

Resolves #1120

### DIFF
--- a/.changes/unreleased/Changed-20260422-204815.yaml
+++ b/.changes/unreleased/Changed-20260422-204815.yaml
@@ -1,0 +1,3 @@
+kind: Changed
+body: 'auth: git-spice will no longer automatically fall back to Git Credential Manager-managed credentials without running ''gs auth login''.'
+time: 2026-04-22T20:48:15.839391-07:00
--- a/.changes/unreleased/Fixed-20260422-170432.yaml
+++ b/.changes/unreleased/Fixed-20260422-170432.yaml
@@ -1,5 +1,5 @@
 kind: Fixed
 body: >-
-  auth: Disable prompting when querying Git Credential Manager (GCM).
-  With prompting enabled, GCM may block the authentication flow.
+  auth: Disable prompting when querying 'git credential',
+  as it may block authentication flow otherwise.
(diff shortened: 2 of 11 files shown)


abhinav:

- #1122 ◀

PR #1126, opened 2026-04-26
branch split: prefill names with branch prefix

### LINKED ISSUE #1119: `git-spice branch split` does not honour spice.branchCreate.prefix
git-spice branch split will take the input of the user directly, ignoring the branchCreate prefix specified in configuration.

I would have imagined that the prefix would also be applied there automatically since we are creating a branch from an existing one?

abhinav:
## Problem

branch split does not currently respect the configured branch prefix.
Unfortunately, there are issues with applying the branch prefix
to branch split unconditionally,
or even as an implicit post-processing step,
as that leaves the current branch name unusable for reuse.

Reusing the current branch name is already a meaningful split operation:
it means "move this existing branch to this split point."
If `feature` is the current branch and `prefix/` is configured,
then always adding the prefix after input changes the meaning:

    current branch: feature
    typed name:     feature
    final name:     prefix/feature

That creates or selects `prefix/feature`
instead of referring to the current branch `feature`.

It's confusing to add prefixing for all branches
except the current branch, with or without exceptions like,
"only if it doesn't already start with the prefix."

## Solution

Rather than deal with confusing or ambiguous behavior,
we seed the interactive branch-name input
with the configured prefix.

With `prefix/` configured, the prompt starts as:

    Branch name: prefix/[cursor]

The prefix is normal editable text.
Users can accept or extend it for new prefixed branches,
or delete it and type the current branch name when they want to reuse it.

`--at` remains literal because it already spells out the exact branch
name to use:

    gs branch split --at HEAD~2:feature

Because, again, we want to allow reuse of the current branch name
without prefixing.

Resolves #1119

### DIFF
--- a/.changes/unreleased/Changed-20260425-101922.yaml
+++ b/.changes/unreleased/Changed-20260425-101922.yaml
@@ -1,0 +1,3 @@
+kind: Changed
+body: 'branch split: Branch name prompts now start with the configured branch creation prefix.'
+time: 2026-04-25T10:19:22.267463-07:00
--- a/branch_split.go
+++ b/branch_split.go
@@ -20,6 +20,8 @@ type branchSplitCmd struct {
 	split.Options
 
 	Branch string `placeholder:"NAME" help:"Branch to split commits of."`
+
+	Prefix string `default:"" config:"branchCreate.prefix" hidden:""`
 }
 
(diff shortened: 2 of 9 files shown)


abhinav:

- #1126 ◀

lowjoel:
thanks!

PR #1129, opened 2026-04-26
fix(sync): Protect unsubmitted merged branches

### LINKED ISSUE #1107: Commit(s) lost on repo-sync with merged branch
Steps to reproduce:

1. Locally:
```
gs bc test
git commit ...
gs ss
git commit ...
```

2. Merge PR.

3. Locally:
```
gs rs
```

Basically, when there is a PR merge and locally added unsubmitted commits, the branch is deleted.

abhinav:
`repo sync` should not delete a branch just because its CR is merged.
If the local branch has commits that were never part of the merged CR,
removing it can discard work that has not reached the forge.
This protects the case where the forge merged an older CR head,
while the local branch moved on afterward.

Use the forge-reported CR head as the deletion interlock.
Merged branches are deleted automatically only when the local head is the
same commit,
or when local Git can prove that the forge head contains the local head.
Otherwise sync keeps the branch in non-interactive mode
and prompts before deletion when interaction is available.

The compact forge status query now returns both state and head SHA,
so the safety check reuses the existing status request without adding
another network call.

Test note:

The renamed-branch sync script now re-submits the branch after rename.
That keeps its original coverage focused on issue #401: deleting the
old tracking ref, `origin/feature1`, when the local branch has been
renamed to `feat1`. Without re-submitting, the local branch contains an
unsubmitted commit and correctly exercises the new #1107 skip path
instead of the renamed-tracking-ref deletion path.

Resolves #1107

### DIFF
--- a/.changes/unreleased/Fixed-20260426-131752.yaml
+++ b/.changes/unreleased/Fixed-20260426-131752.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'repo sync: Prevent branches with unsubmitted local commits from being deleted after their change request is merged.'
+time: 2026-04-26T13:17:52.841836-07:00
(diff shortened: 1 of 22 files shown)


abhinav:

- #1129 ◀

Issue #413, opened 2024-09-18
Allow cross-repo pull requests to be opened


patjakdev:
A common workflow for contributing to open source projects in Github consists of the following steps:

1. Create a fork of the repo that you intend to contribute to
2. Push a branch into your fork containing the changes you wish to merge upstream
3. Open a pull request in the upstream repo which references the branch in your fork

Right now, it doesn't appear that there's a way to push branches from `git-spice` into a fork and open a pull request in another repo. In particular, the `RepositoryID` of the `CreatePullRequestInput` that's passed to Github [here](https://github.com/abhinav/git-spice/blob/8c7ef2c31dac4a3386cff7d2abbc89f950a7aec3/internal/forge/github/submit.go#L23) is always the same as the repo where the branches are pushed.

I'm not sure what it would take to implement this feature. I suppose you'd need to add some config that specifies which repository PRs should be opened in and default it to the same as the one where branches will be pushed.

I'm willing to contribute myself if you'll point me in the right direction.

patjakdev:
I see you have a [similar problem](https://github.com/abhinav/git-spice/blob/main/CONTRIBUTING.md#stacking-changes) in this project 🙂.

In my case, I do have permission to open PRs in the upstream repo, so I'd really like to be able to take advantage of the awesomeness that is git-spice for my contributions.

abhinav:
Hey, @patjakdev.

Unfortunately, this is a [limitation](https://abhinav.github.io/git-spice/guide/limits/#write-access-required) of stacking workflows with GitHub: you can only stack PRs on a repository that you can push branches to.

The workflow where you push a branch to a fork, and create a PR upstream only works if you have a single PR in the stack: the base branch is upstream:main, and the PR head is fork:feature1.
As soon as you have >1 PRs, it doesn't work. E.g. PR 1 is upstream:main..fork:feature1, but PR 2 is fork:feature1..fork:feature2, so the second PR will be created on the fork itself. There's no way to work around this with GitHub; you can't say "upstream:feature1..fork:feature2" without being able to push a branch to upstream.

There are a couple workaround workflows:

- Use git-spice only to manage branches pushed to your fork without creating PRs ([spice.submit.publish=false](https://abhinav.github.io/git-spice/cli/config/#spicesubmitpublish)), manually open a PR upstream for the bottom-most PR in the stack, and when that lands, open a PR for the next one above it, and so on.
[...]

PR #1140, opened 2026-05-02
ui: Add support for prompt footers

abhinav:
Some prompts need active guidance that is more prominent
than the faint description text.
For example, repo init needs to print a clear message
in fork mode (#413) about the degraded experience.

Add a means of extending prompts with a footer.
The footer, if present, renders after the description
and before the error message, separated by a blank line.

Select fields can now provide a footer via a callback
which is able to inspect the currently selected option.


abhinav:

- #1140 ◀

PR #1149, opened 2026-05-12
repo sync: don't restack by default

### LINKED ISSUE #1134: repo sync: do not restack after merge without --restack
Right now, when `gs repo sync` runs, merged branches automatically cause the branches right above them to get restacked, but nothing else. This is incorrect behavior. 

The correct behavior is for the merged branches to be deleted, and for the branches right above them to be re-targeted onto main without restacking — just updating the metadata.

If a restack is required, the user must pass the --restack flag.

NB: This is a significant change in existing behavior even if we consider it a bug/incorrect.
This will need to be advertised clearly, and will likely affect a number of the tests.

kimjune01:
## Summary

- Adds a `SkipRestack` field to `delete.Request`, defaulting to `false` to preserve behavior for explicit branch deletion
- `repo sync` now sets `SkipRestack = !opts.Restack`, so deleted branches are not restacked unless `--restack` is passed
- Adds a golden test (`repo_sync_no_restack.txt`) verifying default no-restack behavior
- Updates existing golden tests to pass `--restack` where rebase behavior was previously expected

Fixes #1134

## Test plan

- [ ] `go test -tags=script -run TestScript/repo_sync` -- all 33 tests pass
- [ ] `go test ./internal/handler/sync/...` passes
- [ ] `go test ./internal/handler/delete/...` passes
- [ ] `repo sync` without `--restack` no longer rebases remaining branches after merge
- [ ] `repo sync --restack` still rebases as before


abhinav:
Thanks, @kimjune01! I was working on this in parallel and fortunately we both went in roughly the same direction so I'm going to push a couple commits here and merge it.

abhinav:
hah, race condition

abhinav:
actually, this isn't quite right. it requires a bit more digging. my initial approach (and this approach) are both not going to work as demonstrated by the test failures.

abhinav:
I'm going to abandon this specific solution path on my end.
Feel free to pursue it further if you'd like but the current version (with or without my changes) isn't valid.

kimjune01:
@abhinav I'll dig in to the test failures and see where I went wrong

kimjune01:
Dug into the test failures. The structural problem is in `CheckRestacked`, not in the delete plumbing. Both your refactor and the original `SkipRestack` shape are correctly wired; they expose a pre-existing gap in restack detection.

## Diagnosis

The headline CI failure is `issue944_special_chars_in_branch_names`. Tracing it:

`main` (rebase-on-delete):
```
INF feature/👨‍💻-developer: moved upstack onto feature/θ-theta   ← real rebase, head moves
INF feature/café: restacked on feature/👨‍💻-developer            ← cascade fires
INF feature/日本語: restacked on feature/café
```

This branch (state-only retarget):
```
INF feature/👨‍💻-developer: retargeted upstack onto feature/θ-theta
INF feature/👨‍💻-developer: branch does not need to be restacked.    ← gap
INF feature/café: branch does not need to be restacked.              ← cascade dies
INF feature/日本語: branch does not need to be restacked.
```

`BranchOnto(SkipRebase: true)` (`internal/spice/onto.go:130-137`) intentionally preserves the recorded `BaseHash` as the OLD base's head, with a comment saying it leaves the branch in a "needs restack" state for later detection. But `CheckRestacked` (`internal/spice/restack.go:143`) only consults `IsAncestor(actualBaseHead, branch.Head)` — it never reads the recorded `BaseHash`. After a state-only retarget, the new base head is trivially an ancestor of `branch.Head` because the branch still carries the orphan commits from the deleted base in its history. So the routine declares "no restack needed" and the cascade dies at the retargeted branch.

The simple cases (`repo_sync_no_restack`, `repo_sync_merged_pr`) pass because they only have one branch above the deleted base — the retarget itself is the only restack the test would notice. Cascades expose the gap.

## Fix

Pushed as `c748bff`. Eight added lines in `CheckRestacked`. Flag NeedsRestack when:

1. recorded `BaseHash != actualBaseHash`, AND
2. recorded `BaseHash` is in `branch.Head`'s ancestry (the orphan commits really are present), AND
3. recorded `BaseHash` is not an ancestor of `actualBaseHash` (rules out plain "base advanced on the same branch", which is harmless and handled by the existing silent-update path).

Condition (2) is what keeps `branch_squash_after_manual_rebase` passing — when the user manually rebases and resets the base backwards, the recorded `BaseHash` is no longer in the branch's history, so we correctly leave it alone.

Also updated the `repo_sync_closed_pr_prompt` golden: the retargeted branch now correctly shows `(needs restack)`, which matches the issue #1134 intent.

## Verification

- `issue944_special_chars_in_branch_names` — was failing on CI, now passes.
- `branch_squash_after_manual_rebase` — protected by condition (2), passes.
- All 33 `repo_sync_*` script tests pass individually.
- `internal/spice/...`, `internal/handler/{sync,restack,checkout,squash}/...` unit tests pass.

Happy to iterate on the predicate or carve the fix differently if you'd rather keep `CheckRestacked` simple and put the detection elsewhere (e.g. a dedicated `IsLazyRetargeted` check on the restack handler side). Just wanted to get the structural diagnosis written down — neither of our delete-shaped diffs was wrong, the detection routine just wasn't built to handle a state-only retarget.

abhinav:
I'm going to close this PR as it has no human in the loop.
I don't mind AI-assisted work, but I'm not going to spend my time getting this into a mergeable state if nobody took the time to make the change.
I was previously working on my own branch of this. As the effort to make that ready is the same as the effort required to get this in, I'm going to focus on my version.

PR #1157, opened 2026-05-13
submit: add `spice.submit.label.addWhen` config

water-sucks:
Currently, `git-spice` adds labels to change requests every time a submission happens. This can break certain workflows that rely on manual removal of labels to kickstart operations.

For example: a label called "skip-ci" exists to stop GitLab CI from starting pipelines on CRs created by `git-spice` by default. The author is expected to remove this label in order to start triggering workflows. `git-spice` currently _always_ adds this label upon submission time, including to existing MRs where the author has manually removed the label, and thus cannot start workflows due to this label getting re-added.

This adds a `spice.submit.label.addWhen` option with two modes:

- `always`: always add labels to CRs at submission time (this is the default/previous behavior),
- `create`: only add labels to CRs when they are immediately created by `git-spice`


doc/src/cli/config.md
```diff
@@ -436,6 +436,23 @@ The value must be a comma-separated list of labels.
 Labels specified with the `-l`/`--label` flags
 will be combined with the configured labels.
 
+<!-- TODO(abhinav): oops, "label" should've been "labels" -->
```
abhinav, line 439: (can't change this at this time. will rename it in the future.)

abhinav:
Thanks!

PR #1158, opened 2026-05-13
submit: add `--fill-from` parameter for stack submission commands

water-sucks:
When submitting stacks, it is impossible to pass custom information for each change request non-interactively. While this is possible for the `gs branch submit` command, passing a title and body for each change request is impossible to do non-interactively.

This is especially useful for facilitating AI agents that automatically submit change requests using `git-spice`, since they can now autofill intelligent CR titles and descriptions instead of relying on default `--fill` output, which isn't that useful.


abhinav:
Thanks, @water-sucks. I'll go over this as soon as I'm able.
I think I might want a different UX for this—here's JSON for all branches doesn't quite match how I'd like for this to work. I'm going to have to think about this a bit more, but regardless, thanks for the change.

water-sucks:
Yeah it's all good! I'd also like to talk about what UX is best for this. This is a patch I wrote on my own that my team is using at work so it's not buried in stone or anything haha

ed-irl:
Relatedly, I have been using an AI based extension of the --fill command for a month or two now; just posted it https://github.com/abhinav/git-spice/pull/1170

PR #1165, opened 2026-05-16
integration: Add state model for integration branch

ed-irl:
Introduce repo-scoped integrationInfo carrying a singleton integration
branch name, the upstream branch, the last pushed hash for
--force-with-lease leasing, and an ordered list of tips with the hashes
recorded at the last rebuild.

Adds Store.Integration / Store.SetIntegration as the public API and
bumps the state storage layout to VersionThree. The version is upgraded
only when an integration is configured, so existing repos that never opt
in keep V1 or V2.

SetRemote preserves integration through remote changes; clearing the
integration drops the version back to its prior level.

Extends the spice.Service Store interface with Integration and
SetIntegration so the service layer can read and write the
configuration. spice.Service.RenameBranch now rewrites any integration
tip that references the renamed branch; rename failures of the
integration update are surfaced as warnings rather than aborting the
primary rename.

Also adds Store.PendingIntegrationRebuild,
Store.SetPendingIntegrationRebuild, and
Store.ClearPendingIntegrationRebuild for persisting a rebuild that was
paused mid-flight (typically by a merge conflict). The handler uses
this to support resuming after the user resolves conflicts in the
worktree.

### DIFF
--- a/.changes/unreleased/Added-20260602-160809.yaml
+++ b/.changes/unreleased/Added-20260602-160809.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'integration: Add storage model for integration branch state (internal)'
+time: 2026-06-02T16:08:09.643553-04:00
--- a/internal/spice/branch.go
+++ b/internal/spice/branch.go
@@ -381,6 +381,40 @@ func (s *Service) RenameBranch(ctx context.Context, oldName, newName string) err
 	}
 	s.log.Debug("Renamed tracked name of branch", "old", oldName, "new", newName)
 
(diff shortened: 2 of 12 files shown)


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165 ◀
                - #1172
                    - #1173
                        - #1174
                            - #1175
                                - #1176
                                    - #1251
                                        - #1252
                                            - #1253
                                                - #1254
                                                    - #1255
                                                        - #1256
                                                            - #1257
                                                                - #1258
                                                                    - #1259
                                                                        - #1260

diegocostares:
I love the idea

PR #1167, opened 2026-05-16
git: add commit message editor package

abhinav:
Implement a new `gitedit` package that coordinates
editing commit messages using the user's Git editor,
replicating the UX of `git commit`.

The package handles the full editor lifecycle:
writing COMMIT_EDITMSG, running prepare-commit-msg
and commit-msg hooks, invoking the editor
with SIGINT/SIGQUIT ignored, and cleaning up
the message based on the configured cleanup mode
(strip, scissors, whitespace, verbatim).

Also adds supporting Repository methods:
Stripspace, DiffTreePatch, and HookRun.

Includes unit tests with mocked Repository,
and integration tests that exercise the package
against real Git repositories, validating
actual comment stripping, hook execution,
verbose diffs, and cleanup mode behavior.

This will be used by `gs commit fixup`
to let users edit commit messages
of non-HEAD commits.

[skip changelog]: no user-facing changes

PR #1168, opened 2026-05-16
commit fixup: Add -e/--edit flag for message editing

abhinav:
Integrate the gitedit.Editor into the fixup handler
so users can modify the commit message when fixing up.

The fixup handler uses plumbing-level CommitTree
rather than git commit, so the editor is invoked
directly via gitedit.Editor.CommitMessage
before passing the result to CommitTree.

The GitRepository interface is extended
with the methods needed by gitedit.Repository
(GitDir, Var, DiffTreePatch, Stripspace, HookRun)
so the handler's repository can be passed through directly.

The Verbose field is named CommitVerbose
with name:"commit-verbose" to avoid conflicting
with the global --verbose flag.

### DIFF
--- a/.changes/unreleased/Added-20260516-142124.yaml
+++ b/.changes/unreleased/Added-20260516-142124.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: >-
+  commit fixup: Add --edit flag to allow editing commit messages.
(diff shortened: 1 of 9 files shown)


abhinav:

- #1016
    - #1167
        - #1168 ◀

PR #1169, opened 2026-05-17
fix(commit-fixup): Preserve dirty edits

abhinav:
When staged fixup changes conflict with the target commit,
commit fixup reports the merge-tree failure before any rebase starts.
Autostash cleanup must restore dirty worktree edits immediately in that
case because there is no rebase continuation that can pop them later.

Keep deferred autostash restoration only for real rebase interruptions
and already-rescued rebase errors. The regression script covers the
staged-conflict path and verifies both the index and dirty worktree edit
survive the failure.


abhinav:

- #1169 ◀

PR #1172, opened 2026-05-19
git: Add Merge, IsClean, and CheckoutNewBranch worktree helpers

ed-irl:
Worktree.Merge supports --no-ff and optional rerere via per-command
-c rerere.enabled=true -c rerere.autoupdate=true flags. On conflict the
merge is automatically aborted (default) and a *MergeConflictError
carrying the refs and unmerged paths is returned, leaving the worktree
at HEAD. Setting MergeOptions.LeaveConflict skips the abort so the
caller can drive a conflict-resolution flow.

Worktree.IsClean reports whether the worktree has uncommitted changes
among tracked files (untracked files are ignored), backed by
git status --porcelain --untracked-files=no.

Worktree.CheckoutNewBranch creates a new branch at an optional start
point in one operation, with a Force flag for git checkout -B semantics.

PR #1173, opened 2026-05-19
integration: Add handler package for integration branch operations

ed-irl:
[skip changelog]: internal handler package; user-facing commands ship in follow-up PRs.

internal/handler/integration provides the business logic for the
integration branch concept: a repo-scoped singleton that combines tips
of multiple tracked branches by sequentially merging them onto trunk.

Handler operations: Create, Delete, AddTip, RemoveTip, Show, Checkout,
Rebuild, and Submit. Checkout switches the worktree to the integration
branch after verifying that it exists. Rebuild merges with --no-ff in
tip order with rerere enabled per-command. On conflict, the merge is
left in the worktree for the user to resolve; the remaining tips and
the captured original tip hashes are persisted via
SetPendingIntegrationRebuild so the next Rebuild invocation resumes
from where it left off. A new *ConflictError carries the conflicting
tip's name and paths. Submit pushes with --force-with-lease against
the last recorded push hash and never calls any forge API.

Two hook methods drive auto behavior from the wrappers:

  - MaybeRebuild detects per-tip hash drift and regenerates when any
    tip has moved.
  - MaybeRebuildAndSubmit additionally pushes when a non-empty
    LastPushedHash signals the user has previously submitted manually.

Both hooks downgrade rebuild conflicts and push failures to warnings so
they cannot poison the outer command.


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165
                - #1172
                    - #1173 ◀
                        - #1174
                            - #1175
                                - #1176
                                    - #1251
                                        - #1252
                                            - #1253
                                                - #1254
                                                    - #1255
                                                        - #1256
                                                            - #1257
                                                                - #1258
                                                                    - #1259
                                                                        - #1260

PR #1175, opened 2026-05-19
integration: Auto-rebuild integration branch after restack

ed-irl:
branch_restack / stack_restack / upstack_restack / repo_restack now
invoke IntegrationHandler.MaybeRebuild after a successful restack. When
integration is configured and a tracked tip's hash has drifted from the
recorded value, the integration branch is regenerated; otherwise it is a
no-op.

A test script exercises the drift-then-rebuild flow and confirms a
second restack with no drift does not rebuild again. merge_wt_test.go
gains explicit git user config so its commits succeed in a clean
environment.

Part of #1244

PR #1174, opened 2026-05-19
integration: Wire gs integration command tree

ed-irl:
Adds the gs integration command tree (alias gs int) backed by the
integration.Handler:

  - show: print configuration and per-tip drift (default subcommand)
  - create: configure the singleton integration branch
  - delete: clear the configuration (leaves the git branch alone)
  - checkout (alias gs intco): switch to the integration branch
  - rebuild: regenerate the branch; --push also submits
  - submit (alias gs ints): force-with-lease push, no PR created
  - tip add/remove/list: manage the tip set
    (add and remove accept multiple branches in one invocation)

A new IntegrationHandler interface in the root package exposes the
handler operations to subcommands and to forthcoming auto-rebuild
hooks. The Kong wiring includes a singleton provider and an
integrationTips completion predictor.

gs branch track now refuses the integration branch name. The
track.Store interface gains Integration(ctx) to support this; the real
state.Store already satisfies it.

End-to-end test scripts cover each command and the cross-cutting
behaviors:

  - integration_create
  - integration_checkout
  - integration_delete
  - integration_rebuild
  - integration_submit
  - integration_tip_add (with variadic args and error cases)
  - integration_tip_rename (gs branch rename rewrites integration tips)
  - integration_branch_cmds_excluded (gs ls hides, gs branch track refuses)
  - integration_conflict_sequential (conflict + git merge --continue +
    gs intrb resume)

Ships the user guide at doc/src/guide/integration.md (wired into
mkdocs nav), regenerated CLI reference, help fixtures, an Added
changelog entry, and the shorthand index.

PR #1176, opened 2026-05-19
integration: Regenerate conflicted derived files on merge

ed-irl:
Adds a git-native `regenerate` merge driver that re-runs the
appropriate generator for derived files when they conflict, rather
than picking a side. Picking a side is unsafe here: many of these
files are only PARTIALLY stochastic. Two branches may have added or
changed real structure (a new flag, test case, mock method) on top of
random ID drift, and a blind take-one silently drops the other side's
structural change. Regenerating against the merged source produces
the correct merged output.

Path mappings, in .githooks/merge-regenerate:

  - internal/forge/shamhub/testdata/**  -> go test -run TestIntegration
                                          ./internal/forge/shamhub/ -update
  - testdata/help/*.txt                 -> go test -run TestHelp . -update
  - doc/includes/cli-*.md               -> mise run generate
  - mocks*.go / mock_*.go / mocks.go    -> mise run generate

Generators that emit many files in one run are deduped per merge via
a tmp marker keyed by the parent PID, so the shamhub test runs at
most once even if all 47 fixture files in its output set conflict.

`mise run setup` registers the driver locally; .gitattributes
declares the path mappings. No post-merge hook is needed: the
driver handles conflicts directly during the merge.

### DIFF
--- a/.changes/unreleased/Added-20260602-160842.yaml
+++ b/.changes/unreleased/Added-20260602-160842.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'integration: Add ''regenerate'' merge driver that re-runs generators for derived files (mocks, CLI ref, shamhub fixtures, help text) when they conflict during merge, instead of picking one side'
+time: 2026-06-02T16:08:42.565148-04:00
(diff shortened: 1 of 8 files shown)


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165
                - #1172
                    - #1173
                        - #1174
                            - #1175
                                - #1176 ◀
                                    - #1251
                                        - #1252
                                            - #1253
                                                - #1254
                                                    - #1255
                                                        - #1256
                                                            - #1257
                                                                - #1258
                                                                    - #1259
                                                                        - #1260

PR #1178, opened 2026-05-19
feat(submodule): track and manage submodule branch associations

ed-irl:
Record which branch each submodule is on when a parent commit is made, and surface those associations in the branch log and through new management commands.

New commands:

- `gs branch submodule list` shows the submodule branch associations recorded for a branch (defaults to the current branch).
- `gs branch submodule repoint <path>` changes the submodule branch associated with the current parent branch, defaulting to the submodule's current branch.

Automatic tracking:

- `branch create`, `commit create`, and `commit amend` record each non-excluded submodule's current branch via a new SubmoduleTracker. Detached-HEAD submodules are skipped, and recording failures only warn.
- `spice.submodule.exclude` lists submodule paths to skip.

Display:

- `gs ls` and `gs ll` render associations as `[path:branch]`; the JSON log adds a `submodules` field.

State:

- branchState gains an additive, optional `submodules` map. UpsertRequest.Submodules merges into existing state: non-empty values set an entry, empty strings remove one, and nil leaves it unchanged.

Git layer:

- Add Worktree.Submodules, SubmoduleCurrentBranch, SubmoduleWorktree, and UpdateSubmodulePointer.
- Fix LocalBranches to resolve the real worktree path for submodules with absorbed gitdirs, where git reports `%(worktreepath)` as the gitdir; previously this caused restacks inside a submodule to be skipped as "checked out in another worktree".

### DIFF
--- a/.changes/unreleased/Added-20260313-074533.yaml
+++ b/.changes/unreleased/Added-20260313-074533.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'submodule: Add `gs branch submodule list` and `gs branch submodule repoint` commands. Submodule branch associations are recorded automatically at commit time and displayed as `[path:branch]` in `gs ls` and `gs ll`.'
+time: 2026-03-13T07:45:33.612939-04:00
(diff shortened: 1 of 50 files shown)


ed-irl:

- #1307
    - #1178 ◀
        - #1179
            - #1180
                - #1181
                    - #1182
                        - #1183

PR #1179, opened 2026-05-19
feat(submodule): add git-layer primitives and transactional/recursive handlers

ed-irl:
Introduce the building blocks for submodule-aware git-spice operations: low-level git inspection/snapshot helpers and a new `internal/handler/submodule` package coordinating them.

- git: add submodule inspection helpers on `Worktree` — `SubmoduleStatus` (branch/detached, HEAD hash, parent gitlink hash, and whether the sub has a gs store), plus the underlying `SubmoduleHead`, `SubmoduleGitlink`, and `SubmoduleHasGsStore`.
- git: add `HeadSnapshot` with `SnapshotHead`/`RestoreHead` to capture and restore a worktree's HEAD (branch or detached), used to make multi-sub operations reversible.
- handler/submodule (Applier): `ApplyAssociations` switches each tracked submodule to the branch recorded for a parent branch, in deterministic order, snapshotting each sub's HEAD and rolling back all switched subs on the first failure. Honors an `Exclude` list and treats unrecorded/untracked branches as no-ops.
- handler/submodule (recurse): `OpenContext` builds the per-submodule git-spice plumbing (worktree, repo, store, service) rooted at a sub, returning `ErrSubmoduleNotInitialized` when the sub lacks a gs store; `ForEachInitializedSubmodule` iterates `.gitmodules`, soft-skipping excluded and uninitialized subs.
- handler/submodule (errors): `DivergedFromRecordError` and `FoldConflictError`, each rendering copy-pasteable remediation hints.
- git tests: replace inline `filepath.ToSlash(filepath.Join(...))` with a `joinSlash` helper.

Part of #1246

PR #1180, opened 2026-05-19
submodule: associations follow checkout, inherit on create, merge on fold

ed-irl:
Make tracked submodule branch associations move with branch operations transactionally, with HEAD rollback on conflict.

- branch checkout (and friends): after switching branches, switch each tracked submodule to the branch recorded for the target. The checkout handler snapshots HEAD first and rolls parent + already-switched submodules back to their original HEADs if the apply fails. Wiring is optional — a nil Submodule applier disables auto-checkout.
- branch create: a fresh child inherits its parent's recorded submodule associations as a baseline (RecordWithInheritance), then overlays the current worktree state so an explicitly-moved submodule wins.
- branch fold: merge the folded child's submodule records into the surviving base (MergeAssociationsForFold). Disjoint and identical records carry over silently; conflicting per-submodule records resolve via the new repeatable '--module-branch=PATH=BRANCH' flag, an interactive prompt, or fail with a FoldConflictError listing all conflicts.
- branch onto: apply the target's submodule associations after BranchOnto runs (it bypasses the checkout handler), snapshotting parent HEAD for rollback on failure.

Adds checkout.GitWorktree.SnapshotHead/RestoreHead, a checkout SubmoduleApplier interface, and unit + test-script coverage for inheritance, checkout apply, and fold merge.

Part of #1246


ed-irl:

- #1307
    - #1178
        - #1179
            - #1180 ◀
                - #1181
                    - #1182
                        - #1183

PR #1181, opened 2026-05-19
submodule: auto-commit and gitlink update on parent commits

ed-irl:
Teach `gs commit create`, `gs commit amend`, and committing
`gs branch create` to include staged submodule work in the parent
commit.

Each tracked submodule with staged changes is committed first, then its
gitlink is updated in the parent commit. Pointer-only changes are
preserved without creating an unnecessary submodule commit. Diverged
submodule state stops the operation before the parent commit is created.

`--module-message=PATH=MSG` supplies a non-interactive message for an
individual submodule commit.


ed-irl:

- #1307
    - #1178
        - #1179
            - #1180
                - #1181 ◀
                    - #1182
                        - #1183

PR #1182, opened 2026-05-19
submodule: gs repo init --recurse-submodules

ed-irl:
Teach `gs repo init` to initialize tracked submodules recursively.

`--recurse-submodules` initializes each eligible submodule as a
git-spice repository. The submodule trunk is resolved from its
`.gitmodules` branch, its HEAD upstream, or the existing trunk guesser.
When the flag is omitted and submodules are present, the command prompts
before recursing; `--no-recurse-submodules` opts out explicitly.

Opting in records `spice.submodule.recurse=true` so later operations
recurse without requiring the flag each time.


ed-irl:

- #1307
    - #1178
        - #1179
            - #1180
                - #1181
                    - #1182 ◀
                        - #1183

PR #1189, opened 2026-05-23
fix TestGitEditor sensitivity to GIT_EDITOR/VISUAL env vars

nitzan-shaked:
## Summary

- `git var GIT_EDITOR` checks `GIT_EDITOR` and `VISUAL` environment variables before `core.editor` config, so tests fail if either is set in the parent shell environment. This fails the test in local envs that have those set (mine, e.g.), which is annoying when one wants to submit a PR.
- Fix: unset both vars with `os.Unsetenv` and restore the original value via `t.Cleanup` (`t.Setenv("FOO", "")` is insufficient: it sets to empty string rather than unsetting, and git still reads it.)

## Test plan

- [x] Run `GIT_EDITOR=true VISUAL=vim go test ./internal/handler/submit/... -run TestGitEditor` before the fix: tests fail.
- [x] Run `GIT_EDITOR=true VISUAL=vim go test ./internal/handler/submit/... -run TestGitEditor` with the fix: all pass.

### DIFF
--- a/internal/handler/submit/form_test.go
+++ b/internal/handler/submit/form_test.go
@@ -17,6 +17,18 @@ func TestGitEditor(t *testing.T) {
 	t.Setenv("HOME", home)
 	t.Setenv("XDG_CONFIG_HOME", filepath.Join(home, ".config"))
 	t.Setenv("GIT_CONFIG_NOSYSTEM", "1")
+	// git var GIT_EDITOR checks GIT_EDITOR and VISUAL before core.editor,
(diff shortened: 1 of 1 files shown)


abhinav (approved):
Thanks!

nitzan-shaked:
> Thanks!

No worries @abhinav ; expect another PR soon, thank me by accepting it :)

Fixed the lint issue (hopefully), resubmitted now.

PR #1190, opened 2026-05-23
fix(nav): reject invalid branch counts

abhinav:
`git-spice up` and `git-spice down` should reject invalid hop counts
with a user-facing error.

When the count is zero or negative, navigation currently falls through
into a blank checkout target and panics.
This change rejects non-positive counts before stack traversal starts,
so both commands fail cleanly instead.

The regression script now covers `0` and `-- -1`,
and the changelog records the new validation contract.


abhinav:

- #1190 ◀

PR #1191, opened 2026-05-23
log: clickable OSC 8 hyperlinks for change IDs in terminal output

nitzan-shaked:
When `git-spice log` output goes to terminal (not JSON, not
piped), change IDs are now wrapped in OSC 8 hyperlink escape
sequences so they are clickable.

Implementation notes:
- Adds a `Hyperlinks` option and a per-item `ChangeURL` field to
  `branchtree.GraphOptions` / `branchtree.Item`. The renderer emits
  the OSC 8 wrapper only when both are present.
- When hyperlinks are enabled, the log command forces the change URL
  to be requested even with `crFormat=id`, so the wrapper has a
  target. URL construction is local (the forge formats it from the
  change ID and repo info); no extra network calls.

### DIFF
--- a/.changes/unreleased/Added-20260523-114404.yaml
+++ b/.changes/unreleased/Added-20260523-114404.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'log: change IDs in terminal output are clickable OSC 8 hyperlinks.'
+time: 2026-05-23T11:44:04.658226+03:00
--- a/.changes/unreleased/Fixed-20260523-112429.yaml
+++ b/.changes/unreleased/Fixed-20260523-112429.yaml
@@ -1,3 +1,3 @@
 kind: Fixed
-body: 'log: Avoid warning about missing remote metadata when listing local-only repositories.'
+body: 'log: Avoid warning about unavailable remote metadata when listing local-only repositories.'
 time: 2026-05-23T11:24:29.04097-07:00
--- a/internal/handler/list/handler.go
+++ b/internal/handler/list/handler.go
@@ -204,7 +204,7 @@ func (h *Handler) ListBranches(ctx context.Context, req *BranchesRequest) (*Bran
 			var ok bool
 			remoteForge, remoteRepoID, ok = forge.MatchRemoteURL(h.Forges, remoteURL)
 			if !ok {
-				return fmt.Errorf("no forge matches remote URL %q", remoteURL)
+				return nil
 			}
 
 			return nil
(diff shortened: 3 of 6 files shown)


abhinav (commented):
Nice!
The project uses lipgloss so I think we can use their [hyperlink functionality](https://pkg.go.dev/charm.land/lipgloss/v2#readme-hyperlinks) to wire this up. This will avoid doing another isatty lookup, and having to pass the hyperlink bool around.

I _think_ it might be as simple as taking a [NewStyle](https://pkg.go.dev/charm.land/lipgloss/v2#NewStyle), adding a hyperlink, and wrapping the output of renderTextWithHighlights with that?

nitzan-shaked:
> Nice! The project uses lipgloss so I think we can use their [hyperlink functionality](https://pkg.go.dev/charm.land/lipgloss/v2#readme-hyperlinks) to wire this up. This will avoid doing another isatty lookup, and having to pass the hyperlink bool around.
> 
> I _think_ it might be as simple as taking a [NewStyle](https://pkg.go.dev/charm.land/lipgloss/v2#NewStyle), adding a hyperlink, and wrapping the output of renderTextWithHighlights with that?

Gotcha. WIll look at that.

@abhinav the formatting is easily done with lipgloss (and then stripped when output is not a tty) but stripping isatty from log.go is not a minor change -- the problem is list/handler.go which wants a remote forge if "wantsChangeURL", before even looking at specific items whether they have a change id or not. So if there's no remote configured, or no forge, there's a WRN -- even if no actual list item has a change id.

This is possible to refactor, of course, but it won't be "just a few lines".

So using lipgloss in tree.go works and the change in tree.go becomes indeed very small. But log.go still needs isatty. I am going to push this change, tell me whether you like it better.

abhinav:
Thanks for looking into that, @nitzan-shaked.
I looked around to refresh my memory, and you're right we seem to be logging a warning.
I think this change is good as-is; the change to lipgloss is easy. 
I'm going to see if that warning might not be an oversight. We shouldn't log a warning just because a remote isn't configured. 

Thanks for the change! I'm going to adjust and merge.

nitzan-shaked:
> Thanks for looking into that, @nitzan-shaked. I looked around to refresh my memory, and you're right we seem to be logging a warning. I think this change is good as-is; the change to lipgloss is easy. I'm going to see if that warning might not be an oversight. We shouldn't log a warning just because a remote isn't configured.
> 
> Thanks for the change! I'm going to adjust and merge.

Oops. I just pushed, this was a race :) I'll stop here, you take it from here. Thank for considering this!

abhinav:
No worries!

Also thanks for catching the log issue. That was a bug. Fixing here: https://github.com/abhinav/git-spice/pull/1192

abhinav (approved):
Thanks!

PR #1192, opened 2026-05-23
log: Suppress missing remote warning

abhinav:
Local-only repositories should be able to use `gs log` and `gs ls`
without seeing warnings for optional remote-backed metadata.

Treat a missing configured remote as an expected degraded mode
for log enrichment.
Other remote configuration errors still warn,
and change URLs continue to fall back to change IDs
when forge metadata is unavailable.

A script regression covers `gs ls --json` on the trunk branch
in a repository with no configured remote.

Discovered in #1191


abhinav:

- #1192 ◀

PR #1193, opened 2026-05-23
chore: Migrate mise task args

abhinav:
mise warns when task run scripts still use Tera argument functions.
Move tasks to usage-backed values.

Manual dry runs covered the default and flag-bearing task paths,
and `mise run build` completes without the deprecation warning.


abhinav:

- #1193 ◀

PR #1195, opened 2026-05-24
fix(top): honor --detach at stack top

abhinav:
`git-spice top --detach` should still detach `HEAD`
when the current branch is already the top-most branch.
Before this change,
`top` returned early as a no-op
and skipped the checkout handler detach path.

This keeps ordinary `top` as a no-op at the stack top
while routing `--detach` through checkout.
The regression script covers both reaching the top branch
and detaching when already there.


abhinav:

- #1195 ◀

PR #1196, opened 2026-05-24
fix(restack): Detect retargeted branch bases

abhinav:
Branches retargeted without rebasing preserve their old base hash
as the upstream boundary for the next explicit restack.

`CheckRestacked` previously treated those branches as already
restacked whenever the new base was an ancestor of the branch head.
That could erase the preserved boundary and prevent explicit restack
from replaying the surviving branch without the old base commit.

Treat that preserved base hash as a restack boundary instead of stale
metadata, so explicit restack rewrites the branch correctly.


abhinav:

- #1196 ◀

PR #1197, opened 2026-05-24
repo sync: Retarget deleted upstacks

### LINKED ISSUE #1134: repo sync: do not restack after merge without --restack

abhinav:
`repo sync` should delete finished branches without silently
rebasing the surviving branch above each deletion.
That keeps sync focused on state and branch cleanup,
while `--restack` remains the explicit command for replaying commits.

Branch deletion now accepts an upstack policy so sync can retarget
surviving upstacks by updating bookkeeping only.
The normal branch delete commands keep their existing rebase behavior.

Script coverage includes the multi-delete stack where `a` and `b`
are deleted from `main -> a -> b -> c`.
After sync, `c` still has its original commit,
is based on `main` in state,
and needs an explicit restack.

Supersedes #1149
Resolves #1134

### DIFF
--- a/.changes/unreleased/Changed-20260501-204224.yaml
+++ b/.changes/unreleased/Changed-20260501-204224.yaml
@@ -1,0 +1,3 @@
+kind: Changed
+body: 'repo sync: Branches above deleted branches are now retargeted without being restacked unless --restack is used.'
+time: 2026-05-01T20:42:24.138949-07:00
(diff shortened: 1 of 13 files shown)


abhinav:

- #1196
    - #1197 ◀

PR #1198, opened 2026-05-25
feat(navigation): Reattach from detached branch

abhinav:
Navigation commands should treat a detached HEAD at exactly one local
branch as a branch-shaped starting point.
This lets `up`, `down`, `top`, and `bottom` move through the stack
without forcing the user to check out the branch manually first.

`top` still reports a no-op when already attached at the top,
but it now honors `--detach` and reattaches when the top branch was
only inferred from detached HEAD.

Targeted tests cover the branch inference boundary and the detached
navigation workflow.


abhinav:

- #1198 ◀

PR #1199, opened 2026-05-25
repo sync: Add restack modes

### LINKED ISSUE #1135: repo sync: --restack scope is incorrect
Right now, `gs repo sync --restack` is the same as `gs repo sync && gs stack restack`: delete merged branches, restack the "current" stack.

There are multiple better interpretations of --restack:

- restack the upstacks of all merged branches
- restack the entire repository (we're in "repo" scope after all)

Following #1134, I think we need to figure out exactly what --restack should do.
Maybe the scope needs to be configurable and not fixed like this.

abhinav:
The previous `repo sync --restack` flag behaved like a bool
and restacked the current stack after sync. That was incorrect.

With #1197, `repo sync` merely retargets base branches of surviving
upstacks, and leaves restacking to the user.

This change teaches --restack to work nicely with that:

- --restack=none (default) does not restack anything
- --restack=aboves restacks only direct upstacks of deleted branches
  (this is the old behavior prior to #1197)
- --restack, --restack=true, --restack=upstack restacks all upstacks of
  deleted branches

The `spice.repoSync.restack` configuration option allows
changing the default restack mode for `repo sync`,

Resolves #1135

### DIFF
--- a/.changes/unreleased/Changed-20260509-102052.yaml
+++ b/.changes/unreleased/Changed-20260509-102052.yaml
@@ -1,0 +1,8 @@
+kind: Changed
+body: >-
+  repo sync:
+  --restack option now runs in 'upstack' mode: all upstack branches of deleted merged branches will be restacked.
+  Override with --restack=aboves to restack only the branches directly above,
+  or --restack=none to disable restacking (this is the default behavior).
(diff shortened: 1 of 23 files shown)


abhinav:

- #1196
    - #1197
        - #1199 ◀

PR #1200, opened 2026-05-25
fix(sync): Preserve deferred restack boundary

abhinav:
Repo sync can retarget an upstack without rebasing it,
then restack that branch later when --restack=aboves is active.
The skipped rebase path must preserve the same upstream
boundary that an immediate BranchOnto rebase would have used.

Store that computed boundary so a later restack does not
replay downstack commits that the branch already contains.
This keeps deferred restacking aligned with the direct rebase path.

This is a follow-up to #1197 and #1199.


abhinav:

- #1200 ◀

PR #1203, opened 2026-05-25
branch onto: Extract onto handler

abhinav:
Move branch and upstack onto orchestration out of the command
implementations and into `internal/handler/onto`.

`spice.Service.BranchOnto` remains the single-branch primitive,
but its request now uses a `BranchOntoMode` enum
instead of the `SkipRebase` boolean.
The zero-value mode preserves the existing rebase behavior,
and retarget-only callers use the explicit mode.

This keeps command files focused on parsing and prompting
while preserving the existing branch onto,
upstack onto,
and branch delete behavior.

### DIFF
--- a/branch_onto.go
+++ b/branch_onto.go
@@ -7,7 +7,7 @@ import (
 
 	"go.abhg.dev/gs/internal/cli"
 	"go.abhg.dev/gs/internal/git"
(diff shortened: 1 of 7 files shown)


abhinav:

- #1203 ◀

PR #1204, opened 2026-05-26
branch onto: Add restack modes

abhinav:
Branches above a moved branch are now retargeted by default
instead of being rebased immediately.

`--restack` keeps an immediate-rebase workflow available
for users who want the affected upstack handled during the move.
Bare `--restack` uses the full upstack mode.
`--restack=aboves` limits the replay
to the direct branches above the moved branch.

`spice.branchOnto.restack` configures the default mode
for `branch onto`.
Explicit CLI flags override the configured value.


abhinav:

- #1203
    - #1204 ◀

PR #1209, opened 2026-05-28
spice.submit.label[.*] -> spice.submit.labels[.*]

abhinav:
`spice.submit.labels` and `spice.submit.labels.addWhen`
are now the preferred configuration keys
for default submit labels and their application policy.

The old singular names continue to resolve as deprecated aliases,
so existing user configuration keeps working
while warning users to move to the plural names.

To aid in this, the config resolver now supports deprecated keys
by tagging fields with `configDeprecated`.
This keeps deprecation policy at the configuration boundary instead
of adding submit-specific lookup code.


abhinav:

- #1209 ◀

PR #1152, opened 2026-05-12
Add 'downstack merge' command to merge stacks

ed-irl:
## Summary

This adds `gs downstack merge` as an experimental command
for merging a stack of Change Requests bottom-up through the forge API.

The command acts as a local merge queue.
It merges the lowest open Change Request first,
waits for that merge to settle,
syncs trunk,
restacks and updates the next branch,
waits for CI checks on the updated Change Request,
and repeats until the requested downstack has been merged.

The command is gated behind the `downstackMerge` experiment
while the merge-queue behavior is still settling.

## Merge flow

For a stack like:

    main <- feature1 <- feature2 <- feature3

running `gs downstack merge` from `feature3` merges:

    feature1, feature2, feature3

After each merge,
repo sync updates trunk,
deletes merged branches,
and retargets their upstacks.
The next queued branch is then restacked,
submitted as an update to its existing Change Request,
checked for passing CI,
and merged.

This keeps squash-merge handling in the merge queue itself,
rather than relying on delete cleanup to repair the stack afterward.

## Test support

ShamHub now supports:

- `shamhub config mergeMethod merge`
- `shamhub config mergeMethod squash`

This lets script tests exercise server-side squash merge behavior
for `gs downstack merge`.

Refs #65

PR #1211, opened 2026-05-30
submit: Share stale-base detection

abhinav:
Move stale-base detection into `spice.FindStaleBases`
so submit validation can use a shared graph and forge-state check
instead of keeping that logic inside the submit handler.

The submit handler still owns its user-facing diagnostics.
It logs each stale local branch/base edge
and returns the existing aggregate error
that asks the user to run `gs repo sync`
or submit with `--force`.

`FindStaleBases` accepts a repository opener
and calls it only after local graph analysis finds candidate bases.
This keeps submit flows that do not need stale-base validation
from requiring a supported forge.

Extracted from #1152


abhinav:

- #1211 ◀

PR #1212, opened 2026-05-31
xec: Bound captured command output

abhinav:
Subprocess failures include captured stderr or stdout in returned
errors when debug logging is disabled.
Very large output should not be retained without a limit before the
command reports its failure.

Limit captured output to the first and last 32 KiB and include a
visible skipped-byte marker between them when output is truncated.
Property tests cover exact reproduction below the limit and the
prefix/suffix contract above it.


abhinav:

- #1212 ◀

PR #1214, opened 2026-05-31
sync: Reuse graph for merged history

abhinav:
Repo sync already loads branch topology before checking supported
forge state.
Merged-branch propagation should use that snapshot instead of
rebuilding topology for every merged branch.

This commit keeps forge queries, deletion planning, and retargeting
behavior unchanged.
It only changes the source of direct upstack data to the existing
`BranchGraph`.

[skip changelog]: user facing impact not measurable


abhinav:

- #1214 ◀

PR #1215, opened 2026-05-31
down: Reuse graph across hops

abhinav:
`gs down N` selects a final checkout target from one stack snapshot.
Rebuilding branch topology for every hop is just a gross inefficiency.

Switch to loading the graph once.
This turns it from an O(M*N) operation to O(M),
where M is the number of branches in the stack
and N is the number of hops.
(We still have to load all branches to build the graph,
but we only do it once instead of N times.)

**Benchmark**

For a degenerate case of a 100-deep stack,
navigating down 50 branches (resetting back to top after each turn),
this is about 38x faster:

| State  | Mean               | Range              |
| ------ | ------------------ | ------------------ |
| Before | 24.445s +/- 1.849s | 22.717s .. 26.396s |
| After  | 633.0ms +/- 81.2ms | 550.3ms .. 712.5ms |

### DIFF
--- a/.changes/unreleased/Changed-20260531-090129.yaml
+++ b/.changes/unreleased/Changed-20260531-090129.yaml
@@ -1,0 +1,3 @@
+kind: Changed
+body: 'down: Improved performance when navigating multiple branches down a stack.'
+time: 2026-05-31T09:01:29.487706-07:00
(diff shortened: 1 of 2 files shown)


abhinav:

- #1215 ◀

Issue #603, opened 2025-02-25
Problems with Git SSH URL when hostname of repo is aliased in SSH config


rturner3:
On my machine, I use multiple GitHub accounts for different purposes (personal vs. work). In order to use my personal GitHub account with some repos and my work account with other repos, I set up some SSH configuration to map the right SSH keys using hostname aliases. Example SSH config:
```
Host githubaccount1
  Hostname github.com
  PreferredAuthentications publickey
  IdentityFile ~/.ssh/githubaccount1_id_ed25519

Host githubaccount2
  Hostname github.com
  PreferredAuthentications publickey
  IdentityFile ~/.ssh/githubaccount2_id_ed25519
```

Then in my Git repo configs, I set the remote URLs to reference these SSH host aliases so that the correct SSH key is used for that repo, e.g. `git@githubaccount1:abhinav/git-spice.git`.

This type of setup doesn't seem to be compatible with git-spice because of an implementation choice to guess the type of Git server being used based on the Git repo URL.

Example:
```
> git remote get-url origin
git@githubaccount1:abhinav/git-spice.git
> gs branch submit
ERR Could not guess repository from remote URL url=git@githubaccount1:abhinav/git-spice.git
ERR Are you sure the remote identifies a supported Git host?
FTL gs: discover CR for my-branch: unsupported Git remote "origin": git@githubaccount1:abhinav/git-spice.git
```

Is there any way git-spice can provide compatibility with this type of Git usage? I realize this is probably not the most common use case, so I'm fine with having to set a little extra config to get this working.

abhinav:
That's an interesting setup. I haven't seen this before, but it makes sense.

I don't know if this will work for sure, but try setting [spice.forge.github.url configuration](https://abhinav.github.io/git-spice/cli/config/#spiceforgegithuburl) to `ssh://$host`.
You'll need to set the [GitHub API URL config](https://abhinav.github.io/git-spice/cli/config/#spiceforgegithubapiurl) to the real API URL.

```
git config spice.forge.github.url ssh://githubaccount1
git config spice.forge.github.apiUrl https://api.github.com
```

That should match the remote URL and treat it as GitHub, which will be enough to extract repository information from the URL, and still send actual requests to GitHub's real API URL.
If authentication doesn't work, you may need to use GitHub CLI's auth method, but I think it'll probably just work.
The resulting PR URLs that it gives you will be garbage (`ssh://githubaccount1/owner/repo`), but it should work otherwise—probably.

---

[...]

PR #1218, opened 2026-06-01
refactor: Add remote forge resolver

abhinav:
Remote repository resolution is shared by submit, sync, auth,
and branch listing flows.
Keeping URL matching policy behind one resolver gives the configurable
forge-kind feature a single boundary to extend.

The resolver preserves URL inference as the default behavior
and keeps the existing unsupported-remote error contract.
Unit tests cover successful forge resolution,
repository ID-only resolution,
and unsupported remotes.

[skip changelog]: no user facing changes

PR #1217, opened 2026-06-01
refactor(forge): Split Git remote parsing

abhinav:
Forge inference and known-forge repository parsing need different
boundaries.
Registry inference must compare a Git remote host against a forge base URL,
but configured forge handling already knows the forge and only needs the
repository path from the same Git remote.

Before this change,
forge discovery passed raw strings through provider implementations:

    f, repoID, ok := forge.MatchRemoteURL(forges, rawRemoteURL)
    repoID, err := f.ParseRemoteURL(rawRemoteURL)

Each provider parsed the full Git remote URL,
checked whether the host matched its configured forge URL,
and extracted the repository owner/name from the URL path.

After this change,
Git remote parsing is explicit and forge base URLs stay separate:

    remoteURL, err := giturl.Parse(rawRemoteURL)
    f, repoID, ok := forge.FromRemoteURL(forges, remoteURL)
    repoID, err := f.ParseRepositoryPath(remoteURL.Path)

This keeps `giturl.URL` scoped to repository remotes,
leaves forge base URL parsing on `net/url`,
and moves path splitting to `forge.SplitRepositoryPath`.

[skip changelog]: no user facing changes

PR #1219, opened 2026-06-01
feat(forge): Add configurable forge kind

### LINKED ISSUE #336: Configurable forge kinds

abhinav:
Some Git remotes use host aliases that are valid for Git transport but do
not identify the forge to git-spice. Add the hidden `spice.forge.kind`
configuration option and `GIT_SPICE_FORGE_KIND` environment variable so users
can select a registered forge without changing their remote URL.

Remote resolution now uses the configured remote URL before Git transport
rewriting for forge detection, while Git operations still use normal Git
transport behavior. Submit, sync, auth selection, and list/log paths share
the same resolver boundary.

Document the setup flow for aliased SSH remotes and cover the config and
environment paths with unit and script tests.

Resolves #336, #603

### DIFF
--- a/.changes/unreleased/Added-20260517-102624.yaml
+++ b/.changes/unreleased/Added-20260517-102624.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'forge: Repositories whose remote host does not identify the forge can now be configured with ''spice.forge.kind''.'
+time: 2026-05-17T10:26:24.393005-07:00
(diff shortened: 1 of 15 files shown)


abhinav:

- #1217
    - #1218
        - #1219 ◀

PR #1221, opened 2026-06-02
sync: detect self-merged stacks pushed directly to trunk

ed-irl:
When a user merges a stack by pushing the stack tip directly to trunk
on the remote (e.g., 'git push origin --force-with-lease HEAD:main'),
the forge often does not report the affected change requests as merged.
On the supported-forge path, repo sync therefore failed to clean up
the stack, leaving each tracked branch with a stale BaseHash that
later restacks would replay against trunk, causing conflicts.

Run the same local-ancestor merge check the unsupported-forge path
already uses as a fallback after the forge API queries: any tracked
branch whose head is reachable from the new trunk hash is treated
as merged, regardless of forge state. Branches with submitted change
requests still attach their ChangeID for downstack history
propagation; unsubmitted branches log "merged into trunk locally".


ed-irl:

- #1307
    - #1221 ◀

PR #1163, opened 2026-05-14
Add review comment commands

ed-irl:
Add `gs review` commands for posting and managing review feedback.

The command group lists remote review threads, posts immediate comments, stages and edits local comments, submits staged comments as one review, edits remote comments, and resolves or unresolves threads.

Staged comments persist by branch until submission. Commands use the optional `ReviewRepository` interfaces, recover forge-owned opaque identifiers from `ListReviewThreads`, and check new comment coordinates against `reviewdiff.Patch` before submission.

PR #1162, opened 2026-05-14
reviewdiff: Parse review comment patches

ed-irl:
Review comment commands need to check postimage anchors before submission, while ShamHub staleness needs to determine whether later revisions delete anchored lines.

Add `reviewdiff.Patch`, backed by `go-gitdiff`, to parse a Git patch once and answer file, line, inclusive-range, and source-line deletion queries. A requested range must fit within one postimage fragment, while deletion queries report only lines actually removed from the source side.

The forge API and provider implementations previously carried by this PR are now provided by #1411 through #1417. Staged comment persistence moves to #1163.

PR #1222, opened 2026-06-02
shamhub: Compute review thread staleness

ed-irl:
ShamHub review threads record the change head in `ReviewThread.CommitHash`. When that revision differs from the current head, compare the two revisions and mark a right-side thread outdated only when the later patch deletes a line in its inclusive range.

Insertions alone do not stale an anchor. File-level and left-side threads retain their stored state because their recorded coordinates do not identify a comparable source range. Explicitly seeded outdated state still takes precedence, and diff or parsing failures propagate to the caller.

This builds on the review patch queries from #1162. The user-facing staleness script is included with the commands in #1163.


ed-irl:

- #1162
    - #1222 ◀
        - #1163
            - #1223
                - #1224

PR #1223, opened 2026-06-02
branch comment: Support file and line-range anchors

ed-irl:
Extend immediate `gs branch comment add` requests from one line to three anchor forms: a whole file, one line, or an inclusive line range.

Parse the command argument into a typed `ReviewThreadRange`, check the complete postimage anchor with `reviewdiff.Patch`, and submit line anchors on the right side of the review diff. File anchors use the zero range defined by the forge review API.

The earlier `--pr` form is removed because review bodies are not returned by `ListReviewThreads` and overlap ordinary change comments. Staged comments retain their existing single-line syntax until their persistence and lifetime contracts are decided separately.

PR #1224, opened 2026-06-02
branch comment: Extend review comment JSON

ed-irl:
`gs branch comment list --json` now reports the review-thread information needed by editor integrations: scope, diff side, inclusive range, reviewed commit, resolution state, and staleness state. The existing compatibility `status` field remains available.

Resolution and staleness are omitted when a forge does not expose those states. File-level records omit line-only fields, range records preserve both inclusive endpoints, and staged replies omit coordinates inherited from their target thread.

The command continues to emit one JSON object per line. Complete deterministic records cover line-range and file-level threads, with focused coverage for unsupported optional state.


ed-irl:

- #1162
    - #1222
        - #1163
            - #1223
                - #1224 ◀

abhinav:
Folding into #1163

PR #1226, opened 2026-06-03
fix: prompt on commit amend after resolving a rebase conflict

### LINKED ISSUE #911: Always prompt user to amend if rebase is in progress
I accidentally used `gs commit amend` today when I was done finishing a rebase conflict instead of `gs rebase continue` which was a bit of a headache to fix!

I see that this is already somewhat handed in that if there is at least one unmerged path the user will be prompted to continue with amending, but I accidentally ran this after fixing all merge conflicts first so didn't get this prompt - appreciate that this is an edge case, but I think it could be useful to still prompt the user to continue with the amend even if all rebase conflicts are resolved?

If not, I should be able to just remember not to `gs commit amend` next time!

mvanhorn:
## Summary

`gs commit amend` now warns before amending when a rebase stopped to resolve a
conflict, even after the conflict has already been staged with `git add`.

## Why this matters

Issue #911: a user finished resolving a rebase conflict (staged every conflicted
file) and then ran `gs commit amend` instead of `gs rebase continue`. Because
the existing guard only fired when there were still unmerged paths, `numUnmerged`
was already `0`, so no prompt appeared and the amend silently rewrote the wrong
commit, corrupting the rebase.

The maintainer's requested discriminator in the issue thread was a way to tell
"the commit we're about to make is a rebase resolution" apart from a deliberate
`edit`/`break` interruption, so the prompt only appears in the resolution case.

This change adds `Worktree.RebaseStopReason`, which classifies the current stop
from the rebase state directory without any new git subprocess:

- last line of `.git/rebase-merge/done` is `pick`/`merge`/`revert`/`fixup`/`squash`
  (or anything unrecognized) -> conflict resolution -> prompt.
- last line is `break` -> deliberate -> no prompt.
- last line is `edit` -> deliberate only if Git wrote the `amend` marker file
  (which it does when the commit applied cleanly and HEAD already holds it).
  An `edit` step whose commit conflicted leaves `edit ...` in `done` but no
  `amend` file, so it is correctly treated as a conflict and prompts.
- missing `done` (e.g. the apply backend) -> defaults to the conflict branch, so
  we err toward a recoverable warning rather than a silent bad amend.

`commit amend` keeps its existing unmerged-paths message and adds a second
message tuned to the resolved-conflict case, preserving the interactive prompt
and non-interactive warning conventions.

## Testing

- `go test ./internal/git/ -run TestWorktree_RebaseStopReason` (pick conflict,
  deliberate edit, break, edit-with-conflict, missing-done).
- `mise run test:script --run commit_amend_rebase_resolution_prompt`
  (resolve+stage a conflict, assert the new prompt appears, "No" aborts).
- `mise run test:script --run commit_amend_rebase_edit_no_prompt`
  (deliberate `edit` stop with a clean tree, assert `gs commit amend` does NOT
  prompt and amends normally).
- `go vet`, `go build`, `gofmt -l` clean; existing `commit_amend_rebase_prompt`
  and `internal/git` suites still pass.

Fixes #911

AI was used for assistance.

### DIFF
--- a/.changes/unreleased/Fixed-20260603-080000.yaml
+++ b/.changes/unreleased/Fixed-20260603-080000.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'commit amend: Prompt before amending when a rebase stopped to resolve a conflict, even after the conflict has been staged.'
+time: 2026-06-03T08:00:00.000000-07:00
--- a/commit_amend.go
+++ b/commit_amend.go
@@ -133,16 +133,17 @@ func (cmd *commitAmendCmd) Run(
 		}
 	}
 
-	// Check if we're in the middle of a rebase with unmerged paths
+	// Check if we're in the middle of a rebase conflict resolution.
 	var rebasing bool
 	if _, err := wt.RebaseState(ctx); err == nil {
 		rebasing = true
 
 		// If we're in the middle of a rebase,
-		// and there are unmerged paths,
+		// and the rebase stopped for conflict resolution,
 		// what the user likely wants is 'git add' and 'git-spice rebase continue'.
 		//
-		// (If there are no unmerged paths, amending is fine.)
+		// (If the rebase stopped deliberately for edit/break,
+		// and there are no unmerged paths, amending is fine.)
 		var numUnmerged int
 		for _, err := range wt.ListFilesPaths(ctx, &git.ListFilesOptions{Unmerged: true}) {
 			if err == nil {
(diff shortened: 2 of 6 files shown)


abhinav (approved):
Thanks!

mvanhorn:
Appreciate you taking this, @abhinav. Prompting on amend after a resolved rebase conflict avoids silently rewriting the wrong commit.

PR #1227, opened 2026-06-06
test: stabilize TestRebase_interactiveRetryPreservesTerminal on Windows

ed-irl:
## Summary

- `TestRebase_interactiveRetryPreservesTerminal` flakes on Windows because the test relies on a goroutine sleeping 20ms before removing a fake `.git/index.lock`. On Windows runners the scheduler doesn't always release the goroutine before the retry path runs — when it doesn't, the test sees the lock still present and fails.
- Replace the goroutine + sleep with atomic lock removal in the first mock call. The lock is removed *before* the mock returns the lock-conflict error, so when the retry path executes there is no timing dependency to lose.

## Before

```go
go func() {
    time.Sleep(20 * time.Millisecond)
    _ = os.Remove(lockPath)
}()

mockExec.EXPECT().
    Run(gomock.Any()).
    DoAndReturn(func(cmd *exec.Cmd) error {
        _, _ = fmt.Fprintln(cmd.Stderr, "fatal: Unable to create '.git/index.lock'")
        return &exec.ExitError{}
    })
```

## After

```go
mockExec.EXPECT().
    Run(gomock.Any()).
    DoAndReturn(func(cmd *exec.Cmd) error {
        // Remove the lock before returning the lock-conflict
        // error so the retry path observes a clean state.
        _ = os.Remove(lockPath)
        _, _ = fmt.Fprintln(cmd.Stderr, "fatal: Unable to create '.git/index.lock'")
        return &exec.ExitError{}
    })
```

No sleep, no goroutine, no race.

## Why this matters

The flake masked CI on multiple unrelated PRs in the current stack (e.g. #1221), wasting workflow runs and obscuring real failures. Fix is independent of any feature work.

## Test plan

- [x] `go test -run TestRebase_interactiveRetryPreservesTerminal -count=10 ./internal/git/` — 10/10 passing locally
- [ ] Verify on Windows CI runner once this lands

### DIFF
--- a/internal/git/rebase_wt_int_test.go
+++ b/internal/git/rebase_wt_int_test.go
@@ -58,14 +58,16 @@ func TestRebase_interactiveRetryPreservesTerminal(t *testing.T) {
 
 	lockPath := filepath.Join(wt.gitDir, "index.lock")
 	require.NoError(t, os.WriteFile(lockPath, nil, 0o644))
-	go func() {
-		time.Sleep(20 * time.Millisecond)
-		_ = os.Remove(lockPath)
-	}()
 
 	mockExec.EXPECT().
 		Run(gomock.Any()).
 		DoAndReturn(func(cmd *exec.Cmd) error {
+			// Remove the lock before returning the lock-conflict
+			// error so the retry path observes a clean state.
+			// A goroutine sleep+remove here flaked on Windows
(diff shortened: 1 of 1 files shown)


ed-irl:

- #1227 ◀
    - #1221

abhinav (approved):

PR #1228, opened 2026-06-07
golangci-lint: fix depguard paths

abhinav:
There appears to be an annoying bug in golangci-lint where,
due to cache-sharing between worktrees (I assume)
lints from other worktrees are reported in the current worktree
with paths using `../../whatever`.
These end up not matching the `path` filtering regex,
and pollute the lint report with false positives.

Work around this by adjusting `path` regexes.

[skip changelog]: not user facing


abhinav:

- #1228 ◀

PR #1230, opened 2026-06-07
forge: Add Bitbucket Data Center support to the bitbucket forge

### LINKED ISSUE #1229: Feat Request: Support self-hosted Bitbucket Data Center / Server
git-spice supports Bitbucket Cloud (bitbucket.org) but not self-hosted Bitbucket Data Center / Server. The two use different REST APIs: Data Center serves `/rest/api/1.0`, while Cloud uses the 2.0 API on api.bitbucket.org. The existing Bitbucket forge targets Cloud and does not work against a Data Center instance.

I'd like to add a `bitbucket-server` forge. Scope:

- Auth with an HTTP access token, sent as a Bearer token.
- Instance URL from `spice.forge.bitbucket-server.url`; API URL derived as `<url>/rest/api/1.0`.
- Forge selected via `spice.forge.kind`, since a self-hosted host isn't identifiable from the remote URL.
- Reuses the existing forge contract, with a separate gateway client for the Data Center API.

Platform constraints (matching Cloud where noted): no PR labels, no assignees, draft state can't change after a PR is created, and no cross-repository PRs.

I have this implemented and validated against a live Data Center instance, and can open a PR. Raising here first per CONTRIBUTING.

kanchev1:
Resolves #1229. Reworked per [the design feedback](https://github.com/abhinav/git-spice/pull/1230#issuecomment-4676576513): one `bitbucket` forge, two API clients — and now a **single repository implementation** over a gateway interface, replacing the previous revision's two parallel implementations.

## What's added

- The `bitbucket` forge now serves both Bitbucket Cloud (REST 2.0) and Bitbucket Data Center / Server (REST 1.0), selecting the client per repository when opening it. No separate forge is registered.
- Product inference: no URL or `bitbucket.org` selects Cloud; any custom URL selects Data Center. `spice.forge.bitbucket.kind` (`cloud`/`datacenter`, alias `server`) overrides the inference, e.g. to stay on the Cloud API behind a custom URL.
- With `spice.forge.kind=bitbucket`, the instance URL is derived from the Git remote URL (context paths included), so a single configuration key suffices for self-hosted setups; `spice.forge.bitbucket.url` always wins when set. Implemented as a small optional `forge.RemoteURLConfigurer` interface used by the remote resolver and `gs auth login` — separable if you'd rather require the explicit URL.
- Both products now fill new pull request descriptions from `PULL_REQUEST_TEMPLATE`-style files on the repository's default branch.

## Design notes

- The gateway seam is laid out as ports and adapters under `internal/gateway/bitbucket`: the package itself defines the product-neutral `Gateway` contract (interface + request/response types), and one adapter package per product — `cloud/` (REST 2.0) and `server/` (REST 1.0) — implements it, each containing its thin REST client alongside the translator. Supporting a future API version means adding an adapter package; the forge is closed against new products.
- `internal/forge/bitbucket` contains **one** `forge.Repository` implementation, written against the `Gateway` contract. The two adapters hide every product difference: pagination models, reviewer identity (account UUIDs vs usernames), Data Center's optimistic-locking versions with one refetch-and-retry on conflict, comment reads via the activity feed, merge vetoes, and draft-support probing. All shared workflow and policy — warnings, checks aggregation, comment counting and filtering, limits, template lookup, metadata — lives once in the repository implementation.
- Optional capabilities use an embedded `UnsupportedGateway` base supplying "unsupported" defaults (the gRPC `UnimplementedXServer` / Gitea `NullDownloader` pattern); the shared repository warns and continues. Today that covers toggling draft status after creation, which Data Center cannot do.
- Two deliberate micro-unifications from merging the implementations, both practically unreachable edge cases: a failed CI check now wins over a pending one on Cloud (previously the first in-progress status short-circuited the scan), and deleting a tracked comment whose persisted metadata lacks the PR ID now errors on both products (previously Data Center silently no-op'd against PR `#0`).

## Compatibility

- Cloud's forge ID, configuration keys, environment variables, stored credentials, and persisted PR metadata are unchanged. The metadata JSON shape is byte-pinned in tests, and the recorded Cloud integration fixtures replay unchanged.
- Behavior change for a released option: a custom `spice.forge.bitbucket.url` now selects the Data Center API (previously it was parsed as Cloud). Noted in a `Changed` changelog entry; `spice.forge.bitbucket.kind=cloud` restores the old behavior.
- The previous revision's `bitbucket-server` forge ID and `BITBUCKET_SERVER_*` variables never shipped and are gone.

## Testing

- A shared repository policy suite runs against a mocked gateway: warnings, checks aggregation, comment counting/filtering, guards, limit defaults, template lookup, metadata construction.
- Per-adapter tests, living in each adapter package, cover the wire mechanics: optimistic-lock conflict retries, activity-feed version recovery, reviewer UUID resolution, default-reviewer merging, draft probing, veto formatting, and both pagination models.
- A cross-product conformance suite next to the contract asserts both adapters produce identical neutral results for equivalent scenarios (states, finds, comments, templates, checks), and pins the documented capability gaps.
- The recorded Cloud integration fixtures replay unchanged, with kind inference, URL derivation (HTTPS, context paths, SSH), and resolver/login wiring covered by unit tests for both products. Data Center behavior is covered by unit tests only; merge and CI-check behavior are not exercised against a live instance.


abhinav:
Thanks @kanchev1, this makes sense to support but I'm reluctant to register a separate forge for this.
Unless I'm misinterpreting how this fits, I think what we'd want is two separate gateway implementations (just clients for the 1.0 and 2.0 REST APIs [^1]) with similar APIs and unifying them in the bitbucket forge implementation, e.g.

We could then do: default is: bitbucket.org with API 2.0, but you can provide a custom URL and we'll switch to API 1.0 (with a configuration option to stay on 2.0 even with a custom URL).

I haven't thought the design through fully, but that's my initial gut reaction. There's a backlog of LLM-assisted PRs so I'm going to come back to this as time permits.

[^1]: They're really clients. I should've called them clients. I'll fix that later maybe.

kanchev1:
Thanks for the feedback, it pointed me in the right direction. I've reworked the PR along those lines. There's no longer a separate forge. It's now just the single `bitbucket` forge backed by two thin REST clients, one for Cloud (2.0) and one for Data Center (1.0), and it picks the right one per repository when the repo is opened.

The configuration is fairly flexible while still doing the sensible thing by default. With no URL set, or `bitbucket.org`, it uses Cloud, and any custom URL switches to Data Center. You can override that inference with `spice.forge.bitbucket.kind` (`cloud` or `datacenter`) if you ever need to. When you set `spice.forge.kind=bitbucket`, it derives the instance URL from the Git remote, including context paths and SSH remotes, so a single config key is enough for self-hosted setups. An explicit `spice.forge.bitbucket.url` still takes precedence when it's set.

I've been running this live against our Data Center instance for day-to-day work, so I can confirm the submit, edit, merge, and comment paths and the URL auto-detection all work well.

I'm also using it together with #1237 (merge-based restack, not merged yet). The two of them together have made git-spice a really useful and helpful tool for me, which is why I wanted to share both upstream in case they're useful to others as well.

Thanks for the great tool. Let me know if you have any remarks on the implementation and I'll adjust or fix anything.

abhinav:
Thanks, @kanchev1, apologies for the delay. I'll follow up on this as soon as I'm able.

abhinav:
Thanks! Will try to make time for #1237 soon as well.

PR #1232, opened 2026-06-08
ui: Add downstack merge progress widget

abhinav:
`gs downstack merge` needs a terminal progress view that can represent a
merge stack without depending on any forge implementation.

Add a `mergeprogress` widget for rendering queue progress as one continuous
bar. Each queue item has its own state, so future merge flows can mark
multiple branches as active at the same time.

The bar uses both glyphs and colors for state:
merged, active, pending, failed, and skipped.
Render tests cover wide and narrow bars, larger stacks, multiple active
items, and event-driven updates.

progress-bar

PR #1233, opened 2026-06-08
ui: Share Bubble Tea model runner

abhinav:
Terminal-only widgets need the same Bubble Tea program setup that forms use.
Without a shared runner, standalone models can miss options such as injected
input, output, window size, TERM, startup messages, and signal handling.

Add `RunOptions` and `RunModel` as the common model execution path.
`Form.Run` now delegates to that path, and `TerminalView.RunModel` fills the
terminal defaults before calling the same runner.

### DIFF
--- a/internal/ui/form.go
+++ b/internal/ui/form.go
@@ -150,8 +150,8 @@ func NewForm(fields ...Field) *Form {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


abhinav:

- #1232
    - #1233 ◀
        - #1234

PR #1234, opened 2026-06-08
merge: Render downstack progress in terminals

abhinav:
`gs downstack merge` now reports merge progress through a typed progress
abstraction instead of writing each loop step directly to the logger.

Terminal views run the stack progress widget so users can see which changes
are pending, active, merged, or failed without repeated stderr chatter.
Non-terminal views keep the existing concise log stream and suppress repeated
polling updates, which preserves script-friendly output.

Focused tests cover non-terminal deduplication and the widget render contract.
The downstack merge script suite continues to verify the non-terminal command
behavior.


abhinav:

- #1232
    - #1233
        - #1234 ◀

PR #1236, opened 2026-06-10
github: Cache user ID lookups

abhinav:
GitHub pull request metadata can resolve the same login
while applying reviewers, assignees, or follow-up edits.
Cache successful user ID lookups on the repository object
so repeated metadata operations can reuse one GraphQL result.

Concurrent cache misses for the same login are coalesced
so parallel callers do not send duplicate `user(login:)` queries.
Tests cover both sequential reviewer-to-assignee reuse
and concurrent lookup reuse against a local HTTP server.


abhinav:

- #1236 ◀

PR #1237, opened 2026-06-10
restack: Add merge-based restack via spice.restack.method

### LINKED ISSUE #454: Merge instead of rebase for restacking
Thanks for the wonderful tool! We have had a great time using it so far.

Currently, we primarily use a merge-based workflow, i.e. if there are any changes down the stack, then rather than rebasing branches upstack, we would like to merge the downstack into upstack to get these changes.

Is there a way to configure git-spice to use merges instead of rebases when restacking?

kanchev1:
Adds an opt-in merge-based restack mode, following the design abhinav sketched in https://github.com/abhinav/git-spice/issues/454#issuecomment-3055229730.

## Why

Rebasing replays every commit of a branch, so a long-lived branch can hit the same conflict on every restack, and every restack needs a force-push. Some servers reject force-pushes. Merging the updated base into the branch instead resolves each conflict once and only appends history (#454).

## What's added

- `spice.restack.method`: `rebase` (default) or `merge`. With `merge`, a restack merges the branch's updated base into the branch. No commits are rewritten, so no force-push is needed.
- `spice.restack.mergeAutoResolve`: `none` (default), `ours` or `theirs`, passed to `git merge -X` for users who want one side to win automatically.
- Top-level `gs continue` and `gs abort`, which resume or cancel an interrupted rebase or merge. `gs rebase continue` and `gs rebase abort` keep working and go through the same code.

This has nothing to do with `gs branch merge` and `gs stack merge`, which merge change requests on the forge. The config docs say so.

## Design

I followed the issue comment where I could:

- Restack is parameterized on a `RestackMethod` enum with rebase as the zero value. `spice.Service` owns the configured method, and the restack and onto handlers read it from there.
- Merge conflicts take the same path as rebase conflicts. A `git.InterruptError` interface covers `RebaseInterruptError` and `MergeInterruptError`, so the existing rescue and continuation machinery resumes either. This is the proposed `RestackInterruptedError`, except that it lives in the git layer.
- `gs continue` and `gs abort` are top-level commands rather than `gs merge continue`. All four commands are thin callers into a new `internal/handler/continuation` package. git-town, Graphite and Aviator also use one top-level continue/abort regardless of the underlying git operation.

The merge commits record conflict resolutions, so a resolved conflict does not come back on later restacks. The merge runs with `rerere.enabled` and `rerere.autoupdate` passed as `-c` options, so nothing is written to git config (see the open question below). Multi-branch merge restacks and `branch onto` in merge mode are wrapped in autostash, so a dirty worktree behaves as it does with rebase.

The default stays `rebase`. Nothing changes unless `spice.restack.method=merge` is set.

One behavior change outside merge mode: `gs abort` and `gs rebase abort` now restore the changes that the interrupted operation autostashed. Before, the `internal autostash-pop` continuation was dropped with the others, so `gs repo restack` on a dirty worktree, a conflict, and `gs rebase abort` silently lost the uncommitted changes (they survived only as a dangling stash commit). This is pre-existing on main. Merge mode hits it on every dirty multi-branch restack, which is how I found it. There is a changelog entry for it.

## Testing

Unit tests cover the merge backend (clean merge, conflict and continue, abort, autostash conflict, a failure that never starts a merge, rerere scoping, `-X` options), method parsing, restack and onto in merge mode, and the continuation handler for rebases and merges. Test scripts run merge restacks end to end: clean, conflict then `gs continue` or `gs abort` (with and without a dirty worktree), unresolved conflicts, auto-resolve, autostash across a multi-branch restack, no recurring conflicts, and `gs rebase continue` during a merge. Another script switches a merge-restacked branch back to rebase, with trunk either advanced or rewritten, and checks that the result is linear with no duplicated trunk commits. The abort fix has its own script for `gs repo restack`.

## Open questions

- rerere: `-c rerere.enabled=true` makes git create `.git/rr-cache` on the first conflict, and git enables rerere by default whenever that directory exists. So after the first conflicting merge restack, rerere is effectively on for the repository. The no-recurring-conflicts behavior comes from the merge commits, not from rerere, so I'm happy to drop the rerere flags if you'd rather avoid the side effect.
- `gs branch onto` and `gs upstack onto` in merge mode merge the new base into the branch, so the old base's commits stay in its history. That's inherent to the method, but the docs don't say so yet. Should it be documented, or should `onto` refuse to run in merge mode?
- Neither the commands nor the config are behind an experiment flag, since they follow the proposed design. Happy to add one if you prefer.
- #1172 also adds `internal/git/merge_wt.go` with a `Worktree.Merge` for the integration-branch work. If it is merged first, I'll rebase onto it and reuse its `Merge`.

## Out of scope

The "rebase if there's no PR, otherwise merge" hybrid policy from the issue thread. It could be a follow-up.

Closes #454


kanchev1:
@abhinav 
## v2: rebased onto main

Same feature and design as v1, rebased onto current main (87c660b2) and reworked so the merge path reuses core git-spice code instead of carrying its own copies. v1 was efc3f1b4 (last head with merges of main: 556cca45).

Changes since v1:

- Rebased onto main and adapted to the upstream refactors: the replay-range logic, the request-struct handlers (#1332), and `errors.AsType`.
- `spice.Service` is now the only owner of the restack method. The restack and onto handlers no longer carry their own field.
- The continue/abort workflow moved into a new `internal/handler/continuation` package. `gs continue`, `gs abort`, `gs rebase continue` and `gs rebase abort` are thin callers into it.
- The merge backend reuses core helpers: `Worktree.Commit` finishes a resolved merge, and the interrupt wrapping and the autostash-pop conflict check are shared with the rebase backend.
- Fixed a pre-existing bug where `gs abort` and `gs rebase abort` dropped the autostashed changes. Details and a changelog entry are in the description.
- New tests: switching a merge-restacked branch back to rebase, abort with a dirty worktree (merge mode and `gs repo restack`), a `git merge` that fails before it starts, and unit tests for the continuation handler.

The open questions for you are in the description.

PR #1239, opened 2026-06-11
help: Use a group key set

abhinav:
In general, sets should be `map[T]struct{}`, not `map[T]bool`.

[skip changelog]: not user-facing


abhinav:

- #1239 ◀

PR #1251, opened 2026-06-12
integration: Add AI-driven conflict resolution

ed-irl:
integration: Add AI-driven conflict resolution

Adds an opt-in fallback that invokes a user-configured shell script
when an integration-rebuild merge conflicts. Rerere and the regenerate
merge driver already handle the mechanical conflicts; this is for
the residue.

The resolver follows a strict JSON protocol on stdout:
  {"assumptions": [...], "questions": [...], "unresolved_files": [...]}
Empty (or assumptions-only) means "all resolved." Questions trigger
an interactive Q&A loop whose pairs persist to .integration_resolution.json
at the repo root. unresolved_files with no questions surfaces as an
error for manual investigation. Non-zero exit or invalid JSON falls
through to the existing conflict-surfacing path.

The resolution file accumulates Q&A across rebuilds so the user
doesn't re-answer for already-resolved (ours, theirs) pairs. Entries
are pruned when branches are untracked, deleted, or removed by sync.

Scoped exclusively to gs integration rebuild and the MaybeRebuild
hook on restack — no impact on branch_merge, restack, or any
user-visible merge. Blast radius: one throwaway branch.

Configured via spice.integration.resolver (script) and
spice.integration.autoResolve (default behavior); --auto-resolve /
--no-auto-resolve override per invocation. Iteration cap is 10 to
prevent runaway loops.

integration: AI-driven auto-resolve for merge conflicts

Adds an auto-resolve path for integration-rebuild merge conflicts.
When configured and enabled, git-spice invokes a user-supplied
script on every merge conflict and, if the script reports the
conflict resolved, stages the files and commits the merge without
intervention.

Shared with the rest of git-spice's script-driven features via the
protocol surface on ed-irl/scriptrun:

  - Uses scriptrun.ResolveResponse for the JSON output schema and
    scriptrun.QAPair for resolution-file entries -- no duplicated
    types.
  - Passes the shared GS_OPERATION (always "integration-rebuild" for
    this feature) / GS_BRANCH (the tip being merged) / GS_BASE (the
    integration branch) env via scriptrun.EnvFor.
  - Iteration cap reads spice.scriptResolve.maxIterations
    (default 10) via Handler.MaxResolveIterations, replacing the
    previously private constant. main.go wires it through.
  - Resolution file lives at .spice/resolutions/integration.json via
    spicedir; the old .integration_resolution.json root-level path
    is gone. Save() auto-creates the parent directory.
  - Resolution-file unknown-field rejection is dropped: per the
[...]

PR #1249, opened 2026-06-12
docs: Add shared protocol guide for script integrations

ed-irl:
Adds doc/src/guide/scripts.md as the single source of truth for the
shared protocol that ties together three soon-to-be-aligned features:

  - spice.message.generator (commit message generation)
  - spice.restack.resolver  (restack auto-resolve)
  - spice.integration.resolver (integration-rebuild auto-resolve)

Documents:

  - Invocation shape (shebang vs sh -c, working dir, args forwarding).
  - Shared env contract: GS_OPERATION, GS_BRANCH, GS_BASE, plus the
    table of all GS_OPERATION values across the three features. Lists
    message-feature-specific extras (GS_MESSAGE_*, GS_TITLE, GS_BODY).
  - JSON output schema (the union of fields across all three features).
  - Assumptions logging, the questions Q&A loop, and termination rules.
  - .spice/resolutions/<feature>.json layout for persistent Q&A.
  - spice.scriptResolve.maxIterations cap.
  - Worked examples and error-handling expectations.

This commit is a prerequisite for the implementation work landing on
ed-irl/scriptrun and the three feature branches above it. Registering
the page in doc/mkdocs.yml's User Guide section.

PR #1188, opened 2026-05-22
scriptrun: Generic runner + shared protocol surface

ed-irl:
Adds the script-execution foundation and the cross-feature protocol
surface that the message generator and both auto-resolvers will use.

Runner (existing): executes a user-configured script either via
'sh -c' or by writing a shebang script to a temp file and exec'ing it.
stdout, stderr, and the exit code are returned without interpretation
so each feature can plug its own handler on top.

Shared protocol (new in this commit):

  - ResolveResponse: the JSON document scripts emit on stdout (the
    union of {title, body, assumptions, questions, unresolved_files}
    fields). ParseResponse with ErrEmptyOutput / InvalidOutputError
    for diagnostics.
  - QAPair: per-question persistent-resolution-file record.
  - Operation enum and EnvFor(operation, branch, base) returning the
    shared GS_OPERATION / GS_BRANCH / GS_BASE env that every script
    receives. Feature-specific env layers on top.

Per-repo scratch directory (new in this commit):

  - internal/spice/spicedir owns <repo-root>/.spice/.
  - Path / EnsureDir for the directory; ResolutionPath /
    EnsureResolutionsDir for per-feature resolution files at
    .spice/resolutions/<feature>.json. Does not modify .gitignore --
    that's a project policy decision left to the user.

Configurable iteration cap (new in this commit):

  - spice.scriptResolve.maxIterations (int, default 10) bounds the
    questions Q&A loop across all features.
  - Config.ScriptResolveMaxIterations() reads it with default+warning
    fallback for invalid values.

Unit tests cover EnvFor, ParseResponse (empty/invalid/valid),
spicedir path + EnsureDir, and ScriptResolveMaxIterations parsing.

Together these are the surfaces that ed-irl/restack-auto-resolve,
ed-irl/integration-auto-resolve, and ed-irl/ai-message-gen will
consume in their respective fixups to collapse the existing
per-feature divergences. See doc/src/guide/scripts.md on
ed-irl/scripts-protocol-guide for the user-facing protocol contract.

PR #1252, opened 2026-06-12
integration: Move regeneration out of merge driver

ed-irl:
integration: Move regeneration out of merge driver

The regenerate merge driver previously ran generators (mise generate,
go test -update) from inside git's merge process. Those generators
side-effect on many files in one pass, and git aborts subsequent merge
steps with 'your local changes would be overwritten' when it sees a
worktree dirtied by an earlier driver invocation. The existing
stage_outputs workaround called git add from inside the driver to
stage the side effects, but git holds the index lock for the duration
of the merge — git add silently failed with rc=128 every time.

This commit moves regeneration out of the merge driver entirely:

  - The driver becomes take-incoming + log: cp the incoming version
    into place, append the path to the file named by
    $GS_INTEGRATION_REGEN_LOG. No side effects.

  - gs runs all tip merges with that env var pointing at a per-merge
    temp file, accumulates the deduplicated path list across the
    rebuild, and invokes .gs/integration-regenerate (a project-level
    executable at the repo root) with the list on stdin. The script
    chooses what to regenerate based on which categories of files
    appear in its input — so a rebuild with no derived conflicts
    pays one process spawn and exits.

  - Whatever the script wrote to the worktree gets folded into the
    last merge commit via git add -A && git commit --amend --no-edit
    --allow-empty.

The contract is well-orthogonalized: gs hands the script an
allow-listed list of paths (only files tagged merge=regenerate in
.gitattributes can appear there), the script returns by writing to
the worktree, gs takes care of the commit-amend wrap-up. Project
generator knowledge stays in the project's script; gs doesn't carry
a dispatch table.

integration regenerate: Regenerate CLI reference


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165
                - #1172
                    - #1173
                        - #1174
                            - #1175
                                - #1176
                                    - #1251
                                        - #1252 ◀
                                            - #1253
                                                - #1254
                                                    - #1255
                                                        - #1256
                                                            - #1257
                                                                - #1258
                                                                    - #1259
                                                                        - #1260

PR #1254, opened 2026-06-12
integration: submit tip downstacks alongside the integration branch

ed-irl:
'gs integration submit' previously only force-pushed the integration
branch. Since the integration branch is the union of its configured
tips, publishing it implies publishing those tips. Otherwise users had
to remember to separately submit each tip stack — easy to forget when
gs ints is meant to be the one-shot "publish the integration" command.

For each configured tip, walk its downstack (trunk-exclusive,
dedup'd), then run SubmitBatch in trunk-first topological order so
bases are pushed before the branches that depend on them. After the
tip-stack batch completes, push the integration branch itself with
its existing force-with-lease semantics.

All the standard submit flags now apply to gs ints (--fill,
--no-publish, --update-only, etc.), so users can keep the previous
"push the branch only" behavior with --no-publish if desired.

### DIFF
--- a/.changes/unreleased/Changed-20260601-111817.yaml
+++ b/.changes/unreleased/Changed-20260601-111817.yaml
@@ -1,0 +1,3 @@
+kind: Changed
+body: '''integration submit'': Also submits the downstack branches of each configured tip, so a single ''gs ints'' publishes the full integration.'
+time: 2026-06-01T11:18:17.074573-04:00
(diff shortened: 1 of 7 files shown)


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165
                - #1172
                    - #1173
                        - #1174
                            - #1175
                                - #1176
                                    - #1251
                                        - #1252
                                            - #1253
                                                - #1254 ◀
                                                    - #1255
                                                        - #1256
                                                            - #1257
                                                                - #1258
                                                                    - #1259
                                                                        - #1260

PR #1255, opened 2026-06-12
integration: surface integration branch and tips in gs ll

ed-irl:
The integration branch was previously invisible to the log commands:
users could only learn about it via 'gs integration show'. That made
it easy to forget which tips were participating, or to overlook the
integration entirely when scanning the stack with 'gs ll'.

Show the integration as a standalone top row of 'gs ls' and 'gs ll'
(rendered with a magenta '◆' marker and a "[integration: N tips]"
annotation), and mark each configured tip in the normal tree with a
trailing "[integration-tip]" badge so the relationship reads at a
glance.

The integration item is wired into the list handler as a synthetic
BranchItem with no base and no aboves, then exposed as a separate
root in the branchtree graph so it renders before the trunk tree.
The JSON output gains an "integration" record on the integration
row and an "integrationTip" boolean on tip rows for programmatic
consumers.

No tree edges are drawn between the integration row and its tips:
that fan-in does not fit cleanly into the existing top-down layout
and the markers already convey the relationship clearly enough.

### DIFF
--- a/.changes/unreleased/Added-20260601-113020.yaml
+++ b/.changes/unreleased/Added-20260601-113020.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: '''log'': Display the configured integration branch as a top row in ''gs ls''/''gs ll'', and mark each configured tip with ''[integration-tip]''.'
+time: 2026-06-01T11:30:20.430333-04:00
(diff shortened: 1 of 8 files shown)


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165
                - #1172
                    - #1173
                        - #1174
                            - #1175
                                - #1176
                                    - #1251
                                        - #1252
                                            - #1253
                                                - #1254
                                                    - #1255 ◀
                                                        - #1256
                                                            - #1257
                                                                - #1258
                                                                    - #1259
                                                                        - #1260

PR #1256, opened 2026-06-12
integration: short-circuit gs branch checkout for the integration

ed-irl:
'gs bco <integration>' previously fell into the generic "branch not
tracked" path: in interactive mode the user was prompted to track the
branch (and accepting would then fail with "cannot track integration
branch"), and in non-interactive mode they got a misleading warning
to run 'gs branch track'. Both are noise — the integration is
deliberately untracked and the user typed its name because they want
to switch to it.

Detect the integration branch in the branchCheckoutCmd up front. If
the requested name matches, emit a single targeted warning that the
branch is throwaway and short-circuit ShouldTrack to return false so
the existing CheckoutBranch path falls through to the plain checkout
without prompting.

This makes 'gs bco <integration>' a viable alias for 'gs intco'
without introducing a separate command path or changing the checkout
handler's contract.

### DIFF
--- a/.changes/unreleased/Added-20260601-113312.yaml
+++ b/.changes/unreleased/Added-20260601-113312.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: '''branch checkout'': Treat the integration branch name as a shortcut to the integration branch, with a one-line throwaway warning instead of an untracked-branch prompt.'
+time: 2026-06-01T11:33:12.704475-04:00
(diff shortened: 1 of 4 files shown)


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165
                - #1172
                    - #1173
                        - #1174
                            - #1175
                                - #1176
                                    - #1251
                                        - #1252
                                            - #1253
                                                - #1254
                                                    - #1255
                                                        - #1256 ◀
                                                            - #1257
                                                                - #1258
                                                                    - #1259
                                                                        - #1260

PR #1257, opened 2026-06-12
integration: gs ss from integration submits all tip stacks

ed-irl:
integration: gs ss from integration submits all tip stacks

'gs ss' is "submit the current stack", but the integration branch is
not part of any stack — it's the union of multiple. Running 'gs ss'
from there previously crashed with "branch not found in store"
because the lookup against the branch graph fails for untracked
branches.

When the current branch matches the configured integration name,
dispatch into an integration-aware path: collect the full stack
(upstack + downstack, trunk excluded) of each configured tip in
trunk-first topological order, dedup branches shared across tips,
hand the result to SubmitBatch, then push the integration branch
itself with its existing force-with-lease semantics.

This is broader than 'gs integration submit' (which only submits the
downstack of each tip); 'gs ss' is the "submit everything in this
stack" command, and from integration that means upstack too.

integration-stack-submit: Regenerate derived files

### DIFF
--- a/.changes/unreleased/Changed-20260601-125125.yaml
+++ b/.changes/unreleased/Changed-20260601-125125.yaml
@@ -1,0 +1,3 @@
+kind: Changed
+body: '''stack submit'': From the integration branch, submits the full stack of every configured tip (upstack + downstack) and then pushes the integration branch.'
+time: 2026-06-01T12:51:25.554189-04:00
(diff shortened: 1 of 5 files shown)


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165
                - #1172
                    - #1173
                        - #1174
                            - #1175
                                - #1176
                                    - #1251
                                        - #1252
                                            - #1253
                                                - #1254
                                                    - #1255
                                                        - #1256
                                                            - #1257 ◀
                                                                - #1258
                                                                    - #1259
                                                                        - #1260

PR #1258, opened 2026-06-12
integration: tip clean prunes subsumed tips

ed-irl:
integration: tip clean prunes subsumed tips

When two configured tips lie on the same stack, merging the lower
one into the integration branch produces no content that the higher
tip's merge doesn't already produce. Keeping both wastes a merge
per rebuild and inflates the resolver-prompt surface for no benefit.

Add 'gs integration tip clean' (alias 'prune') that walks each tip's
upstack looking for another configured tip. When one is found, the
lower tip is subsumed: remove it and log which tip subsumes it.
With 'low -> middle -> high' all configured, the prune cascades:
'low' is reported as subsumed by 'middle' (the nearest tip up the
stack), and 'middle' as subsumed by 'high'. Both come out and only
'high' is left.

A clean run with nothing to prune reports 'No subsumed tips to
prune' instead of being silent.

The new command is layered at the CLI: it reads tips via the
existing Show, computes subsumption against the branch graph, and
removes via RemoveTip. The integration handler interface stays
unchanged.

integration-tip-clean: Regenerate derived files

### DIFF
--- a/.changes/unreleased/Added-20260601-125454.yaml
+++ b/.changes/unreleased/Added-20260601-125454.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: '''integration tip clean'' (alias ''prune''): Remove tips whose upstack chain already contains another configured tip.'
+time: 2026-06-01T12:54:54.098874-04:00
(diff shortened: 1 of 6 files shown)


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165
                - #1172
                    - #1173
                        - #1174
                            - #1175
                                - #1176
                                    - #1251
                                        - #1252
                                            - #1253
                                                - #1254
                                                    - #1255
                                                        - #1256
                                                            - #1257
                                                                - #1258 ◀
                                                                    - #1259
                                                                        - #1260

PR #1259, opened 2026-06-12
integration: tip advance moves tips to upstack leaves

ed-irl:
Extending a stack above a configured tip leaves the tip "behind" —
the integration branch keeps merging the old tip and ignores the new
top of the stack. The workaround so far has been 'tip remove old'
plus 'tip add new', which gets tedious when several tips drift
together.

Add 'gs integration tip advance' that walks the upstack tree of each
configured tip via graph.Tops, swaps the tip for its leaves, and
dedups across tips. Linear case ('low -> middle -> high' with tip
'low') collapses to a single replacement: 'low' becomes 'high'.
Forked case ('base -> {fork-a, fork-b}' with tip 'base') expands to
both leaves. Already-at-the-top tips are left alone with no log
spam. An empty advance reports "No tips to advance".

The optional positional argument scopes the operation to a subset of
configured tips, so 'tip advance fork-a' walks only that tip.
Unrecognized names abort before any mutation so we never leave the
tip list half-updated.

### DIFF
--- a/.changes/unreleased/Added-20260601-125928.yaml
+++ b/.changes/unreleased/Added-20260601-125928.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: '''integration tip advance'': Move each configured tip to the leaves of its upstack, optionally scoped to specific tips.'
+time: 2026-06-01T12:59:28.111325-04:00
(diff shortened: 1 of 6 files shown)


ed-irl:

- #1307
    - #1249
        - #1188
            - #1165
                - #1172
                    - #1173
                        - #1174
                            - #1175
                                - #1176
                                    - #1251
                                        - #1252
                                            - #1253
                                                - #1254
                                                    - #1255
                                                        - #1256
                                                            - #1257
                                                                - #1258
                                                                    - #1259 ◀
                                                                        - #1260

PR #1261, opened 2026-06-12
test/shamhub: Scope bare-repo git commands with --git-dir

ed-irl:
git 2.50 no longer auto-discovers a bare repository from cwd alone, so
running commands like `git config`, `git show-ref`, and `git rev-parse`
via WithDir(repoDir(...)) fails with "fatal: not in a git directory".
This broke every shamhub-backed script test.

Add a `(*ShamHub).gitCmd(ctx, owner, repo, args...)` helper that prefixes
the command with `--git-dir <bare-repo>` and replace each WithDir-on-
bare-repo call site with it. Behavior is unchanged on older git
versions; tests now pass on 2.50+.

### DIFF
--- a/internal/forge/shamhub/change.go
+++ b/internal/forge/shamhub/change.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"go.abhg.dev/gs/internal/forge"
-	"go.abhg.dev/gs/internal/xec"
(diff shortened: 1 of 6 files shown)


ed-irl:

- #1261 ◀

PR #1265, opened 2026-06-13
github: Implement ChecksByChange with per-run detail and checks URL

ed-irl:
Implements the GitHub side of the per-change checks API: rollup
state, per-run name/state/url for both CheckRun and StatusContext
nodes, and the PR's checks summary page URL.

  - Per-PR GraphQL query over PullRequest.commits(last:1).
    StatusCheckRollup gives us the rollup state and the contexts
    connection (capped at 50 per PR; pagination not yet needed).
  - rollupFromStatusCheckRollupState collapses the StatusState enum
    to the upstream rollup taxonomy. Empty input -> ChecksNone
    distinguishes 'no checks' from 'all passed'. Unknown states
    fall back to ChecksFailed.
  - nativeCheckRunState lower-cases CheckRun.conclusion when status
    is COMPLETED, otherwise CheckRun.status. nativeStatusContextState
    lower-cases StatusContext.state. Forge-native vocabulary is
    preserved in runs[] so future UI surfaces can distinguish
    neutral/cancelled/skipped/timed_out without an upstream API
    change.
  - PullRequest.permalink + '/checks' provides the click target.
  - ChangeChecksState (merge handler) gets a small refactor onto
    the new rollup helper without changing its existing 'no rollup
    -> passed' semantics.

ChecksByChange issues one query per PR. Batched-via-aliases would
be sharper for tall stacks but is awkward in shurcooL/githubv4;
the per-PR queries are cheap relative to the rest of gs ll.

Integration fixtures regenerate on the next forgetest -update
pass (real GH credentials).

PR #1264, opened 2026-06-13
shamhub: Implement ChecksByChange with full rollup + runs + URL seeding

ed-irl:
Implements the per-change checks API for the in-process forge so
script tests and forge-suite tests can exercise the new payload
without going to a real CI system.

  - shamChange grows ChecksRuns and ChecksURL fields alongside the
    existing ChecksState. Backward compatible: SetChangeChecksState
    still works for tests that only seed a rollup.
  - SetChangeChecks(owner, repo, n, *forge.ChangeChecks) seeds the
    full structure; ChecksByChange(owner, repo, ns) reads it back,
    treating zero-value ChecksState as ChecksNone (no checks
    configured/reported, distinct from passing).
  - New POST /{owner}/{repo}/change/checks-by-change endpoint
    mirrors the comment-counts batch shape. Returns one nil slot
    per unknown change so callers can correlate by index.
  - forgeRepository.ChecksByChange dispatches to the endpoint and
    parses the response back into forge.ChangeChecks.
  - parseChecksState learns 'none' so the wire round-trip matches
    the new ChecksState value added on the schema branch.
  - shamhub test-script CLI gains 'shamhub set-checks <owner/repo>
    <pr> <json>' for seeding rich payloads from .txt scripts.

PR #1263, opened 2026-06-13
forge: Add ChecksByChange and ChecksNone for per-change checks output

ed-irl:
Lays the schema foundation for emitting per-change CI/build checks
in the gs ll JSON output. The actual rollup mappings + runs[] data
for each forge land in follow-up branches on this stack.

Adds:
  - forge.ChecksNone, a new ChecksState value for 'no checks
    configured/reported' (distinct from ChecksPassed). The card
    indicator hides on none; pending/failed surface visually.
  - encoding.TextMarshaler/Unmarshaler on ChecksState so the state
    serializes as 'pending'|'passed'|'failed'|'none' in JSON.
  - forge.ChangeChecks{Rollup, Runs, URL}: rollup drives the card,
    Runs preserves forge-native state strings for fine-grained
    surfaces (e.g. a future diff-view preamble), URL is the
    forge's per-change checks page.
  - forge.CheckRun{Name, State, URL}: per-check detail. State stays
    a string (forge-native vocabulary) so the upstream API doesn't
    have to enumerate every forge's taxonomy.
  - Repository.ChecksByChange(ids), a batch method mirroring
    CommentCountsByChange.

Each forge's ChecksByChange is stubbed to return a slice of nils
(equivalent to 'no data' under the contract) so the schema commit
builds standalone; real implementations land per-forge in the
follow-ups.

PR #1266, opened 2026-06-13
gitlab: Implement ChecksByChange with per-job pipeline detail

ed-irl:
Implements the GitLab side of the per-change checks API: rollup
state from the MR's head pipeline, per-job name/state/url from the
pipeline's job list, and the pipeline's web URL as the click target.

  - gateway: Pipeline grows ID and WebURL fields; new Job type
    (name + status + web_url); new PipelineJobsList API method
    wrapping /projects/:id/pipelines/:pipeline_id/jobs.
  - rollupFromPipelineStatus collapses GitLab's pipeline status
    enum to the upstream rollup taxonomy. Canceled/manual/skipped
    are user-initiated non-failures and roll up as passed (the
    operator's choice not to run them isn't a failure). Unknown
    statuses default to pending — GitLab adds new statuses over
    time and pending is the safer default.
  - pipelineState (the merge handler's path) keeps its 'no
    pipeline -> passed' semantics; ChecksByChange uses ChecksNone
    in that case so the indicator can distinguish.
  - ChecksByChange issues MergeRequestGet + PipelineJobsList per
    MR. Jobs are reported in pipeline order (stages first, then
    job order within stages); GitLab's job.status is already
    lowercase so it flows straight through as forge-native state.

Integration fixtures regenerate on the next forgetest -update
pass (real GitLab credentials).

### DIFF
--- a/internal/forge/gitlab/checks_report.go
+++ b/internal/forge/gitlab/checks_report.go
@@ -2,18 +2,93 @@ package gitlab
 
 import (
 	"context"
+	"fmt"
 
 	"go.abhg.dev/gs/internal/forge"
+	"go.abhg.dev/gs/internal/gateway/gitlab"
 )
 
-// ChecksByChange reports per-change rolled-up and per-run check state
-// for each of the given changes.
+// rollupFromPipelineStatus collapses a GitLab pipeline status into the
+// forge rollup taxonomy. Canceled/manual/skipped pipelines are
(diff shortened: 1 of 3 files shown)


ed-irl:

- #1307
    - #1263
        - #1264
            - #1265
                - #1266 ◀
                    - #1267
                        - #1268
                            - #1269

PR #1267, opened 2026-06-13
bitbucket: Implement ChecksByChange with per-status build detail

ed-irl:
Implements the Bitbucket side of the per-change checks API: rollup
state from the PR head commit's build statuses, plus per-status
name/state/url forwarded as runs[].

  - gateway: CommitStatus grows Key, Name, and URL fields. Existing
    callers and fixtures keep working because the new fields are
    omitempty.
  - rollupFromCommitStatuses collapses Bitbucket build-status
    states to the upstream rollup taxonomy. Failure wins over
    pending wins over passed. Empty list means no CI configured
    -> ChecksNone. Stopped statuses are operator-cancelled
    non-failures and roll up as passed (changed from the merge
    handler's aggregateStatuses, which treated Stopped as Failed;
    that path is preserved for backwards compatibility but no
    longer informs the richer API).
  - ChecksByChange issues getPullRequest + CommitStatusList per
    PR. State is lowercased to match the spec's forge-native
    convention; Name falls back to Key when the build-status
    provider didn't supply a human name.
  - aggregateStatuses (the merge handler's path) keeps its
    pre-existing semantics and is now scoped narrowly so the
    refactor doesn't change merge behavior.

Integration fixtures regenerate on the next forgetest -update
pass (real Bitbucket credentials).

### DIFF
--- a/internal/forge/bitbucket/checks_report.go
+++ b/internal/forge/bitbucket/checks_report.go
@@ -2,18 +2,104 @@ package bitbucket
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"go.abhg.dev/gs/internal/forge"
+	"go.abhg.dev/gs/internal/gateway/bitbucket"
 )
 
-// ChecksByChange reports per-change rolled-up and per-run check state
-// for each of the given changes.
+// rollupFromCommitStatuses collapses a Bitbucket build-status list into
+// the forge rollup taxonomy.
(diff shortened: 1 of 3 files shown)


ed-irl:

- #1307
    - #1263
        - #1264
            - #1265
                - #1266
                    - #1267 ◀
                        - #1268
                            - #1269

Issue #65, opened 2024-05-22
Idea: Local Merge Queue


abhinav:
For repositories that use Merge for PRs, merging a stack of PRs is very easy.
Given a stack `main -> feature1 -> feature2`, you can merge feature1, and then feature2.
No additional work is needed.

For repositories that use Squash or Rebase for PRs, the workflow is:
merge feature1, then rebase feature2 on new main, push, wait for CI, press merge again.
This is annoying.

It's worth exploring a merge command that acts as a local merge queue:
it merges the bottom-most PR, waits for the merge (e.g. if a GitHub Merge Queue is in use), rebases the next PR, pushes, merges, and so on.

Obviously, it stops as soon as the build fails (maybe with a build retry option for known flaky builds), or if there's a conflict rebasing, but otherwise it would be useful. e.g. hypothetically,

```
> gs downstack merge
Merging feature1 (#42)
Waiting for Merge Queue...
Rebasing feature2 (#43)
Waiting for builds...
Merging feature2 (#43)
Waiting for Merge Queue...
All PRs merged!
```

matthiasr:
If feasible, it would be ideal if this could run in the background without depending on the local workspace state. That way, I could set a stack to merge, while continuing to work on other things (either on top of or in parallel to this stack).

jarviliam:
Running in the background would be ideal but it may need to run in the foreground to report errors when it's unable to merge due to failing pr checks/conditions.
Having a flag that controls if the merge queue is background tasked or not seems to be a good idea 👍

matthiasr:
I don't need it to run in the background, I would open another terminal for parallel work, as long as that doesn't mess up the merge process.

abhinav:
Let me prefix this by saying that this feature is currently in just-an-idea phase, so implementation details are super undecided.

The way I imagined this working, it would need access to *a* working copy of the repository so that it could restack upstack branches after the bottom-most PR merged. So it would need to hold onto the working tree while it's waiting to merge everything. *However*, it doesn't have to take over the working tree that you're working inside. We would be able to use the [git-worktree](https://git-scm.com/docs/git-worktree) command to get a secondary tree. So depending on how the implementation looked, the command could either always create a temporary working tree for rebasing or make it opt-in/opt-out with a flag[^1].

[^1]: Populating a new working tree can take long for really really large monorepos.

matthiasr:
[...]

PR #1268, opened 2026-06-13
list: Emit per-change forge checks via --cr-checks JSON flag

ed-irl:
Wires the per-change checks data from each forge into 'gs log
--json' so the extension and other tooling can consume rolled-up
state and per-run detail.

  - internal/handler/list: new IncludeChecks flag in the request
    bitmask; new Checks field on BranchItem; batch-loaded via
    Repository.ChecksByChange alongside change state and comment
    counts in the existing errgroup.
  - log.go: new --cr-checks/spice.log.crChecks flag (negatable,
    default false) on commit log / commit log long. Honored by
    the JSON presenter; the graph presenter ignores it for now
    (UI surface lives in the extension per ext #65).
  - JSON schema: change objects gain an optional 'checks' field
    with {rollup, runs[], url}, mirroring the spec at ext #100.
    Rollup is one of pending/passed/failed/none; runs[] preserves
    forge-native state strings.
  - testdata/script/log_checks.txt: end-to-end script covering
    failed/passed/none rollup states, per-run detail, --no-cr-checks
    omission, and the spice.log.crChecks config path.

Tracks irl-llc/git-spice-code-extension#100.

### DIFF
--- a/doc/includes/cli-reference.md
+++ b/doc/includes/cli-reference.md
@@ -312,9 +312,11 @@ Before merging, the stack is checked for branches
 whose base PR was already merged on the forge.
 Use --no-branch-check to skip this validation.
 
-Before each merge, waits for CI checks to pass.
-Use --build-timeout to configure the maximum wait
-before failing if checks are not ready.
+Before each merge, waits for merge readiness:
+the forge must observe the pushed head
+and report that the CR is ready to merge.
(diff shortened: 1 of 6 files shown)


ed-irl:

- #1307
    - #1263
        - #1264
            - #1265
                - #1266
                    - #1267
                        - #1268 ◀
                            - #1269

PR #1273, opened 2026-06-13
ui: Add scroll-region model runner

abhinav:
Merge progress needs a terminal mode where a Bubble Tea model can keep
redrawing while ordinary command output continues to write to the same
stream. That mode must preserve resize behavior without using the
alternate screen and without turning progress into repeated log lines.

The standard Bubble Tea renderer owns cursor movement and resize setup.
That works for full-screen programs,
but it collides with command logs when both write to the terminal.
Disabling the renderer avoids that contention,
but it also means the command must supply the startup size
and resize events itself.

Add a scroll-region runner to the shared UI layer.
The runner wraps any compact Bubble Tea model,
draws it into reserved bottom rows with DECSTBM margins,
and leaves normal writes to scroll above those rows.
The wrapped model still receives `tea.WindowSizeMsg` values,
but the terminal-size source now belongs to the runner.

`TerminalView` also carries the command `sigstack.Stack`.
That lets the scroll-region resize watcher share signal handling
with the rest of the command instead of registering a separate
`os/signal` handler.

The reserved height is a min/max range.
The minimum prevents the reserved region from bouncing
when a model briefly renders fewer rows.
The maximum leaves space for log output
as future widgets grow more detailed.

### DIFF
--- a/internal/ui/form.go
+++ b/internal/ui/form.go
@@ -10,6 +10,8 @@ import (
 
 	"charm.land/bubbles/v2/key"
 	tea "charm.land/bubbletea/v2"
+	"go.abhg.dev/gs/internal/sigstack"
+	"go.abhg.dev/gs/internal/ui/scrollregion"
 )
 
 // FormKeyMap defines the key bindings for a form.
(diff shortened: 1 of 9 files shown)


abhinav:

- #1273 ◀
    - #1274

PR #1274, opened 2026-06-13
merge: Run progress widget in scroll region

abhinav:
We had to revert #1234 because the widget used Bubble Tea's standard rendering,
which conflicted with the command log output.

This takes advantage of #1273, which added scroll-region reservation.
When `gs downstack merge` uses the terminal widget now,
ordinary logs continue to print above the reserved region
while the widget redraws in the bottom rows.

The configured height range is four to eight rows,
which fits the current compact status view and leaves room
for short future detail sections.

https://github.com/user-attachments/assets/8da855a9-8f95-406a-aeb9-814834b88138

Reapply #1234
Reverts eb0f912a239eaf271a189823b21d9b6ae80e8252


abhinav:

- #1273
    - #1274 ◀

PR #1275, opened 2026-06-13
merge: Wait for pushed head before CI

abhinav:
After a downstack merge restacks and submits an upstack branch,
the forge can lag behind the local branch state.
The existing merge loop asked for the aggregate check state immediately.
That allowed stale passing checks from the previous change head,
or a temporary no-rollup state,
to satisfy the wait before CI for the pushed head had run.

Wait until the forge reports the same change head
that the merge loop will pass to `MergeChange`.
Only after that head is visible do we ask whether checks passed.
This keeps `gs downstack merge` from merging a restacked branch
based on check state that belonged to an older head.

The regression covers the stale-head ordering directly:
checks are not queried until the forge reports the newly pushed head.

### DIFF
--- a/internal/handler/merge/handler.go
+++ b/internal/handler/merge/handler.go
@@ -442,6 +442,16 @@ func (e *mergePlanExecutor) execute(
 
 		// CI is checked after retargeting and restacking
 		// so each merge waits on the exact change being merged.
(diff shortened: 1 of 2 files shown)


abhinav:

- #1275 ◀

Issue #1243, opened 2026-06-12
scriptrun: generic script-runner utility for script-driven recipes


ed-irl-codebot:
## Problem

Several planned features need to run user-configured shell scripts and
capture their output — AI message generation, integration conflict
resolution, and restack conflict resolution all want the same
"run a recipe, read what it produced" primitive. Without a shared
utility each would reimplement script execution, output capture, and
exit-code handling slightly differently.

## Proposal

Add a small, single-responsibility `scriptrun` package for executing
user-configured scripts and capturing their output.

- Scripts may be plain shell snippets (run via `sh -c`) or
  self-contained executables starting with a shebang (written to a temp
  file and executed directly).
- stdout and stderr are captured independently and returned with the
  exit code.
- The package does not interpret output — that is the caller's job —
  keeping it reusable across different output protocols (commit
  messages, conflict resolutions, etc.).
- A non-zero exit is reported in the result rather than as an error, so
  callers can distinguish "script ran and disagreed" from "script could
  not be run at all."

This is foundational infrastructure for the AI message generation and
integration/restack resolver features.

## Tracked by

- #1188 — `scriptrun` script-runner utility

abhinav:
scriptrun was merged in #1276. it's being used in the merge feature for spice.merge.command. marking as done. 
(PR linked above is not the one that originally implemented scriptrun.)

PR #1276, opened 2026-06-13
scriptrun: Add generic script-runner utility

abhinav:
Introduces a small, single-responsibility package for executing
user-configured shell scripts and capturing their output.

Scripts may be either plain shell snippets (passed to 'sh -c') or
self-contained executables starting with a shebang line (written to
a temp file and executed directly).

stdout and stderr are captured independently and returned alongside
the exit code. The package does not interpret script output; that's
the caller's responsibility. This keeps scriptrun reusable across
different output protocols (commit messages, conflict resolutions,
etc).

A non-zero exit is reported in the RunResult rather than as an error,
so callers can distinguish 'script ran and disagreed' from 'script
could not be run at all.'

Part of #1243

This intentionally narrows PR #1188 to only the core scriptrun
functionality described in issue #1243: the generic script runner and
its focused runner tests. It excludes resolver.go, env.go, parser
errors, .spice storage, and script-resolve configuration.

[skip changelog]: internal package extracted from PR #1188.

### DIFF
--- a/internal/scriptrun/scriptrun.go
+++ b/internal/scriptrun/scriptrun.go
@@ -34,6 +34,7 @@ import (
 	"os"
 	"strings"
 
+	"go.abhg.dev/gs/internal/must"
 	"go.abhg.dev/gs/internal/silog"
 	"go.abhg.dev/gs/internal/xec"
 )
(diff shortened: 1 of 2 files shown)


abhinav:

- #1276 ◀

PR #1278, opened 2026-06-14
mergeprogress: Animate active segments

abhinav:
Active merge progress segments now show a small moving marker
so long-running merge waits have visible progress without changing
merge state updates or forge behavior.

The marker stays inside the active segment and falls back to the
existing static glyph when the segment is too narrow.
Widget callers can disable animation for deterministic snapshots.

[skip changelog]: feature unreleased


abhinav:

- #1278 ◀

PR #1280, opened 2026-06-14
diffmap: Add diff position mapper

abhinav:
Comment workflows need to translate working-tree line numbers into
diff positions before a forge can accept inline comment requests.

The diff mapper parses unified diff hunks into a small mapping API that
callers can use without depending on any forge implementation.
Keeping this package independent lets forge comment support and CLI
state management adopt the mapper separately.

Extracted from #1162


abhinav:

- #1280 ◀

abhinav:
Closing because the retained diff functionality has been redesigned as `reviewdiff.Patch` and moved into #1162.

PR #1279, opened 2026-06-14
feat(branch): Add branch merge command

abhinav:
Users can merge one branch that is configured directly on trunk:

    gs branch merge
    gs branch merge --branch feature

The command uses the existing forge-backed merge loop for the selected branch. That keeps merge confirmation, CI waiting, merge method selection, build timeout handling, and one-item progress reporting consistent with downstack merge.

Branches configured on another branch are rejected with guidance to use `gs downstack merge --branch <name>` for stack-level merge behavior. The branch command deliberately exposes only the shared merge options, so downstack-only behavior such as `--no-wait` and stale-base bypass stays off this command.

Focused scripts cover current-branch defaulting, explicit branch selection with failed CI, confirmation, and configured-base rejection.

PR #1282, opened 2026-06-14
refactor(merge): Add merge queue scheduler

abhinav:
Stack-level merge commands need to schedule branch graphs
where multiple branches can become eligible
after a shared base merges.
That scheduling policy should be testable
without depending on forge APIs,
local Git state,
logging,
or terminal rendering.

Teach `internal/graph.Toposort` to return cycle errors
that preserve the cycle path.
`internal/mergequeue` uses that topological order
as its scheduling baseline,
so queued bases become eligible before their upstacks
even when callers provide items out of dependency order.

Introduce `internal/mergequeue` as the generic queue boundary.
The scheduler accepts opaque item IDs and base IDs,
rejects duplicate IDs and cycles,
and reports item failures with `ItemError`.
Non-fail-fast execution skips only the failed item upstack
and continues independent branches;
fail-fast execution skips all remaining pending work.

The merge handler now adapts `mergeItem` values into queue items
and lets `mergePlanExecutor` implement the queue executor directly.
Preparation,
CI waits,
forge merge calls,
and trunk sync stay in the existing merge executor
so merge-loop output still flows through progress events
instead of direct logging.

Queue observers report both failed and skipped items.
The progress bridge uses that to mark failed branches in the widget
while preserving the more specific failure message
emitted by the merge step that returned the error.

### DIFF
--- a/internal/graph/topo.go
+++ b/internal/graph/topo.go
@@ -1,39 +1,98 @@
 package graph
 
-import "go.abhg.dev/gs/internal/must"
+import (
+	"fmt"
+	"strings"
+
+	"go.abhg.dev/gs/internal/must"
+)
 
 // Toposort performs a topological sort of the given nodes.
 // parent returns the parent of a node, or false if the node doesn't have one.
 //
 // Values returned by parents MUST be in nodes.
-// The graph MUST NOT have a cycle.
+// If the graph has a cycle,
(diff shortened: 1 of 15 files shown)


abhinav:

- #1278
    - #1279
        - #1282 ◀
            - #1283

PR #1285, opened 2026-06-14
fix(mergeprogress): Avoid flickering between frames

abhinav:
Animated merge progress updates
were redrawing every reserved row on each tick.
That made the widget prone to flicker
while ordinary merge output scrolled above it.

Use ultraviolet to render the reserved region
so unchanged cells are left alone between frames.
The scroll-region wrapper still owns most of the work;
ultraviolet only diffs the model rows inside that reserved region.

mergeprogress-widget


abhinav:

- #1285 ◀

PR #1287, opened 2026-06-15
forge: Return structured change checks

abhinav:
Merge readiness needs provider check details instead of only an
aggregate pass, pending, or fail value.
Expose `Repository.ChangeChecks` so every forge can return the check
shape available from its existing status-query path.


abhinav:

- #1287 ◀

PR #1290, opened 2026-06-20
fix: Forward stdin to shell aliases

abhinav:
Shell command aliases can wrap interactive git-spice commands
or commands that read piped input,
but the alias subprocess previously inherited stdout and stderr only.
That left commands such as `git-spice bco -n`
with no input stream when the alias was invoked through git-spice.

Forward stdin to the shell command subprocess
so configured shell aliases behave like ordinary shell commands.


abhinav:

- #1290 ◀

PR #1291, opened 2026-06-20
fix(branch track): Reject non-branch operands

abhinav:
`gs branch track HEAD` previously accepted `HEAD` because branch tracking peeled the operand as a revision and wrote a `HEAD` branch record.

Branch tracking now checks that the requested name exists as a local branch before base inference or state updates. Revision names that do not name local branches are rejected instead.

Regression coverage verifies that `HEAD` is rejected and leaves only the trunk branch tracked.


abhinav:

- #1291 ◀

PR #1292, opened 2026-06-20
fix(branch checkout): Allow dry-run worktree picks

abhinav:
The interactive branch checkout prompt already lets `--detach` choose branches checked out in another worktree because no branch switch happens in the current worktree. `--dry-run` has the same no-switch contract: it only prints the selected branch name.

Before this change, `git-spice branch checkout -n` displayed those branches but rejected the selection as `unknown branch` because the prompt item was disabled. Dry-run mode now leaves those items selectable and still returns before mutating the worktree.

The worktree prompt regression covers selecting a checked-out branch with `--dry-run` and verifies that HEAD is unchanged.

Verification:

- `mise run test:script --run branch_checkout_prompt_detach_worktrees`
- `go test ./internal/handler/checkout`
- `mise run lint`
- `mise run build`


abhinav:

- #1292 ◀

PR #1294, opened 2026-06-20
fix(nav): Suppress dry-run restack warnings

### LINKED ISSUE #1289: nav: `--dry-run` prints restack warning
`gs down -n` printed a warning when the branch below was the bottom-most branch and needed to be restacked. this is a bug. dry run should just report the name.

abhinav:
Navigation dry runs are meant to report the resolved branch name without
running checkout-side validation or mutations.

Checking restack state before the dry-run return made `gs down -n` print
a restack warning when the target branch needed restack, even though the
command was only answering which branch it would select.

Dry-run checkout now returns before restack verification, branch recovery,
or tracking prompts.

Resolves #1289


abhinav:

- #1294 ◀

PR #1293, opened 2026-06-20
forge: Add mergeability API

abhinav:
## Summary

- Add `ChangeMergeability` to the forge repository contract.
- Keep CI/check details under `ChangeChecks` instead of overloading check status as merge readiness.
- Provide initial provider implementations that return the zero-value unknown result so provider translation branches can define their own policies separately.

## Testing

- Worker validation previously covered focused forge tests and review repairs for this branch.

[skip changelog]: internal forge API boundary for unreleased merge work.

PR #1295, opened 2026-06-20
test(forgetest): Add mergeability scenarios

abhinav:
Mergeability integration coverage needs to prove the forge-reported
state that real changes produce.

Test against Ready, Draft, and Conflicts scenarios
that the shared suite can create and inspect across providers.

We're holding off on testing CI failures
as that requires much more machinery to set up,
and the API calls are still the same.


abhinav:

- #1287
    - #1293
        - #1295 ◀
            - #1296
            - #1297
                - #1299
                    - #1300
            - #1298

PR #1297, opened 2026-06-20
github: Translate mergeability state

abhinav:
GitHub exposes two relevant merge readiness signals.

Detailed merge state status carries branch protection, draft,
and policy decisions.
The older mergeable field is useful as a fallback for conflict information.

PR #1296, opened 2026-06-20
bitbucket: Translate mergeability state

abhinav:
Bitbucket reports merge readiness through pull request mergeability and
conflict fields.
Map those fields into the shared forge result.
Merge commands can then treat Bitbucket like other forges.

The shared mergeability integration scenario remains skipped for
Bitbucket with an explicit TODO.
The project does not currently have a recording account for those
fixtures, so unit coverage exercises the provider translation contract.

PR #1298, opened 2026-06-20
gitlab: Translate mergeability state

abhinav:
GitLab exposes merge readiness through detailed merge status on merge requests.
Map that provider-specific status into the shared forge result.
Merge commands can then wait on that same contract across forges.

GitLab reports some conflicting fast-forward-only merge requests as
`need_rebase` with `has_conflicts` set.
Decode that extra field so those merge requests report conflicts.
They should not report a plain behind state.


abhinav:

- #1287
    - #1293
        - #1295
            - #1297
                - #1298 ◀
                    - #1296

PR #1301, opened 2026-06-22
fix(ui): Print logs through Tea renderer

abhinav:
## Why

The first merge progress integration used a reserved terminal scroll
region for the widget.
That made the widget feel stable while regular log output scrolled
elsewhere.

That approach also meant git-spice owned terminal margins,
resize handling,
and a second render loop beside Bubble Tea and ultraviolet.
The code kept fighting the renderer that already owned the screen.
In practice,
small changes in log output or progress width could leave stale terminal
cells behind.

A representative corrupted frame looked like this:

    ■■■■■■====–=■■■■■■====----------------
    1   pend11   waiting: 1   pending: 8

That failure mode was not just cosmetic.
It meant the progress widget could show impossible counts or a broken
bar while a merge operation was still running.

## Alternatives Tried

We tried to keep the scroll-region model alive with narrower fixes.
Those experiments included hand-rendering the affected rows,
clearing rows before repaint,
resetting renderer state,
and custom row-level diffing.

Some variants caused flickering.
Some variants reduced corruption.
None gave us a small implementation
that clearly fit Bubble Tea's rendering model.
The strongest signal from those experiments was that owning a second
terminal renderer made every future widget feature riskier.

## New Design

Anything reliable needed Bubble Tea and log output to coordinate.
If we're going to coordinate anyway,
might as well use Bubble Tea's rendering path for log output.

This change stops reserving a scroll region for merge progress.
The merge progress widget stays inside Bubble Tea's normal renderer.
Instead, log output is printed through Bubble Tea's `tea.Println`.
These are printed as messages above the active widget,
the widget moves down one line, and continues rendering as usual.

mergeprogress-outputwriter-vhs

### Implementation

Output goes through `ui.OutputWriter`.
`ui.OutputWriter` writes directly to the underlying stream
when no Bubble Tea program is active.
While a program is active,
`ui.OutputWriter` line-buffers output lines
and sends them to Bubble Tea as print messages.
Bubble Tea then decides how to place those lines
relative to the current model frame.

`RunModel` unwraps `ui.OutputWriter`
before constructing the Bubble Tea program,
so Bubble Tea still renders the model to the real terminal stream.
Only ordinary writes are redirected through the active program.

When the program exits,
any buffered partial line is written
directly to the underlying writer.

This is less clever than the scroll-region runner,
but it lets Bubble Tea own the terminal state during a program run.
When the merge widget gets more complex in the near future,
this lowers our maintenance burden.


abhinav:

- #1301 ◀

PR #1302, opened 2026-06-23
anchor: Make restack and log anchor-aware

ed-irl:
Anchors are per-worktree trunks: graph roots disconnected from the canonical trunk. This makes restack and log surface them.

- `gs repo restack` now also restacks stacks rooted at anchors, which an upstack-from-trunk walk never reaches. `repo restack -w` from inside an anchor worktree restacks that worktree's own anchor-rooted stack.
- `gs ll` injects an anchor display node for another worktree's anchor-rooted stack, tagged `[root anchor]` or `[internal anchor]`. JSON output gains `anchor` and `anchorKind` fields.
- Checking out a branch already in use by another worktree now fails with a legible `BranchInUseError` naming the worktree, instead of git's raw fatal.

Adds unit tests, test scripts, and a changelog entry.

PR #1187, opened 2026-05-22
feat: add worktree management commands and per-worktree trunks

ed-irl:
Add `gs worktree create` and `gs worktree list` for managing Git worktrees.

`worktree create` sets up a per-worktree trunk by default: a local branch
(named after the worktree directory, or set with `--trunk`) that tracks the
same remote ref as the canonical trunk. Stacks in different worktrees are
based on their own trunk, so `repo sync` and restacks never contend on a
single shared trunk checkout. `--no-trunk` keeps the detached-HEAD behavior,
and `-b` creates a tracked branch stacked on the worktree trunk.

Worktree trunks are recorded in a new, additive state key and treated as
graph roots: `IsTrunk` and `BranchGraph` traversals stop at any trunk, and
`TrunkFor` resolves the trunk that applies to a given worktree. Trunk sync
now pulls the canonical remote ref into the worktree's local trunk.

Add `-w/--worktree` to `log short` and `repo restack` to scope operations to
stacks with a branch checked out in the current worktree.

Part of #1247

PR #1303, opened 2026-06-23
anchor rm: remove an anchor worktree and dissolve its anchor

ed-irl:
Adds `gs anchor rm <path>`, the counterpart to `anchor create`/`track`, for tearing down an anchor worktree once its work is done.

The command resolves the path to an anchor (matching the registry's recorded worktree path, then falling back to the branch checked out there), then:

- Refuses to remove the current worktree up front, since Git won't remove the working tree it's run from.
- Removes the worktree first — the only step that can fail for reasons outside our control (a dirty tree without `--force`) — so a failure leaves all state untouched and the command is safe to retry.
- Retargets the anchor's direct children onto its base (the pinned branch for an internal anchor, else the canonical trunk) in retarget-only mode, leaving them needing a restack rather than risking a conflict.
- Deletes the anchor's pointer branch and unregisters the anchor.

`--force` removes a worktree with uncommitted changes.

Includes a test script covering self-removal refusal, the dirty-tree no-op-on-failure guarantee, successful removal with child retargeting, and the non-anchor-path error, plus help golden files.

### DIFF
--- a/anchor.go
+++ b/anchor.go
@@ -8,4 +8,5 @@ type anchorCmd struct {
 	Create anchorCreateCmd `cmd:"" aliases:"c" help:"Create a new worktree anchored at a branch"`
 	List   anchorListCmd   `cmd:"" aliases:"ls" help:"List anchors and their worktrees"`
 	Track  anchorTrackCmd  `cmd:"" aliases:"tr" help:"Track an existing worktree as an anchor"`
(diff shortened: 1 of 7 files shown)


ed-irl:

- #1307
    - #1187
        - #1302
            - #1303 ◀
                - #1304
                    - #1305

ed-irl:
Closing this: the worktree/anchor approach didn't pan out in multi-agent use (parking branches weren't usable as submit bases, and the store-tracked worktree relationships were fragile under concurrent writes). I'm withdrawing the series and will revisit worktree support with a different design.

PR #1304, opened 2026-06-23
repo: Add exclusive mode (park/restore/exclusive)

ed-irl:
Adds an exclusive mode that hands the whole repository to a single process so it can reorganize the stack without contending with worktrees owned by other processes.

New commands:

- `gs repo park` records every linked worktree in a durable manifest (in spice state) and removes the worktree directories, leaving the branches intact and reachable from the primary checkout. Dirty worktrees are refused unless `--force`. The manifest is written before any removal, so an interrupted park resumes by re-running.
- `gs repo restore` re-creates the parked worktrees at their branches' tips and clears the marker. It is idempotent and resumable. If a parked branch was deleted out-of-band, restore leaves the rest in place, stays in exclusive mode, and explains how to recover; `--forget PATH` discards a worktree whose branch is gone on purpose.
- `gs repo exclusive -- <cmd>` wraps a command in park/restore, always restoring afterward even if the command fails.

While parked, `gs anchor create` is refused to avoid racing the parking process. Each parked commit is pinned under `refs/gs-park/` so it survives GC until the worktree is recovered.

State gains `Park`/`Unpark`/`InExclusiveMode`/`ParkedWorktrees` backed by an additive, optional `exclusive` store key, and git gains `DeleteRef`. Covered by unit tests and test scripts for the round-trip, detached HEAD, dirty refusal, crash recovery, branch-gone wedge, forget, and the exclusive wrapper.

### DIFF
--- a/.changes/unreleased/Added-20260620-063403.yaml
+++ b/.changes/unreleased/Added-20260620-063403.yaml
@@ -1,0 +1,3 @@
+kind: Added
+body: 'repo: Add exclusive mode: `gs repo park` records and removes linked worktrees so one process owns the whole repository, `gs repo restore` brings them back, and `gs repo exclusive -- <cmd>` wraps a command in park/restore. While parked, `gs anchor create` is refused.'
+time: 2026-06-20T06:34:03.614351-04:00
--- a/anchor_create.go
+++ b/anchor_create.go
@@ -63,6 +63,13 @@ func (cmd *anchorCreateCmd) Run(
 			"there is no anchor branch to name")
 	}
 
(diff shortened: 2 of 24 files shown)


ed-irl:

- #1307
    - #1187
        - #1302
            - #1303
                - #1304 ◀
                    - #1305

ed-irl:
Closing this: the worktree/anchor approach didn't pan out in multi-agent use (parking branches weren't usable as submit bases, and the store-tracked worktree relationships were fragile under concurrent writes). I'm withdrawing the series and will revisit worktree support with a different design.

PR #1306, opened 2026-06-23
commit: support multiple -m/--message flags

### LINKED ISSUE #1288: commit: multiple -m flags
git commit allows multiple -m flags to be supplied. we should do the same for all commands that create commits.

mvanhorn:
`git commit` accepts multiple `-m`/`--message` flags and joins them as
separate paragraphs of the commit message. This teaches git-spice's
commit-creating commands to do the same.

## Why

git-spice positions its commit commands as drop-in shortcuts for
`git commit` (`gs commit create` == `git commit` + restack). But the
`-m` flag was bound to a scalar string, so `gs commit create -m subject
-m body` silently dropped everything but the last message, diverging
from Git and surprising users who rely on the multi-`-m` idiom to write
a subject plus body without opening an editor. Issue #1288 asks that all
commands that create commits mirror Git's repeated-`-m` behavior.

## What changed

- `gs commit create`, `gs commit amend`, and `gs branch create` now bind
  `-m`/`--message` to a repeatable `[]string`. Each occurrence becomes a
  separate paragraph, mirroring `git commit -m ... -m ...`.
- `sep:"none"` is set on the flag so a message containing commas is kept
  verbatim (kong's default comma-splitting would otherwise break messages
  like `-m "Fix bug, take two"`).
- `git.CommitRequest` gains a `Messages []string` field that emits one
  `-m`/value arg pair per element. The existing scalar `Message string`
  field is left untouched, so all other callers keep working. A single
  `-m` produces git arguments identical to before.

## Behavior

```
gs commit create -m "Subject line" -m "Body paragraph."
```

produces a commit whose message has the subject and body as separate
paragraphs, exactly like Git.

## Testing

- New unit test `TestWorktree_Commit_multipleMessages` in
  `internal/git/commit_wt_test.go` covering the multi-paragraph happy path
  and single-message parity.
- New test script `testdata/script/commit_multiple_messages.txt`
  exercising the end-to-end CLI path.
- Regenerated help goldens (`testdata/help/*.txt`) and the CLI reference
  (`doc/includes/cli-reference.md`) for the updated help text.
- Added an `Added` changelog entry.

Fixes #1288

### DIFF
--- a/.changes/unreleased/Added-20260623-050407.yaml
+++ b/.changes/unreleased/Added-20260623-050407.yaml
@@ -1,0 +1,6 @@
+kind: Added
+body: >-
+  commit create, commit amend, branch create: The -m/--message flag may now
+  be passed multiple times to add multiple paragraphs to the commit message,
+  matching 'git commit'.
+time: 2026-06-23T05:04:12.000000-07:00
--- a/branch_create.go
+++ b/branch_create.go
@@ -28,9 +28,9 @@ type branchCreateCmd struct {
 	Below  bool   `help:"Place the branch below the target branch and restack its upstack"`
 	Target string `short:"t" placeholder:"BRANCH" help:"Branch to create the new branch above/below"`
 
-	All         bool   `short:"a" help:"Automatically stage modified and deleted files"`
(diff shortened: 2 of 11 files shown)


abhinav (approved):
Thanks!

PR #1307, opened 2026-06-24
restack: Add Options parameter to high-level restack entry points

ed-irl:
`RestackBranch`, `RestackStack`, and `RestackDownstack` now take an `*restack.Options` argument, exposing a per-invocation `AutoResolve *bool` override that is threaded through to `Request.AutoResolve`. A nil `Options` (or nil `AutoResolve`) preserves the existing default behavior.

All callers in the merge, sync, and CLI restack commands pass nil for now; the `RestackHandler` interfaces and generated mocks are updated to match the new signatures.

### DIFF
--- a/branch_restack.go
+++ b/branch_restack.go
@@ -32,5 +32,5 @@ func (cmd *branchRestackCmd) AfterApply(ctx context.Context, wt *git.Worktree) e
 }
 
 func (cmd *branchRestackCmd) Run(ctx context.Context, handler RestackHandler) error {
-	return handler.RestackBranch(ctx, cmd.Branch)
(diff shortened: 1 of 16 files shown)


ed-irl:

- #1307 ◀
    - #1262
        - #1308

abhinav:
extracted just the options plumbing into https://github.com/abhinav/git-spice/pull/1332, moving Options into Request structs. the pattern we've been using with handler options is that they define CLI flags and configs, and the outer request object defines the general runtime parameters.

abhinav:
Closing because current restack handlers use typed request structs on main, superseding this options-parameter API.

PR #1308, opened 2026-06-24
submit: Refuse to overwrite commits pushed to a branch by someone else

ed-irl:
`gs branch submit` previously force-pushed with a lease against the remote-tracking ref, which advances on every fetch and so could silently overwrite commits a maintainer (or anyone else) added to a branch after git-spice last pushed it.

Submit now leases against the commit git-spice itself last recorded pushing. Because that baseline only changes when git-spice pushes, it survives local restacks while still catching foreign commits. Before pushing, submit classifies the remote:

- **Advanced** — the remote gained commits on top of our last push. Submit refuses, lists the offending commits (now including author email), and points to `gs branch sync`.
- **Diverged** — the remote was rewritten away from our last push. Submit refuses since it can't tell what a force-push would discard.
- **No baseline** — no record of the last push and the remote differs. Submit refuses rather than guess.
- **Safe / fast-forward / up-to-date** — push proceeds, leasing against the recorded baseline.

`--force` overrides all of these.

Supporting changes:

- `git.ListCommitsDetails` now also reports each commit's author email (`%ae`), used when listing foreign commits.
- `gs branch sync --rebase` records the integrated remote head as the new baseline. A conflict-interrupted sync resumes via a hidden `--record-pushed` continuation that records the baseline without re-running the rebase.
- `branch split` carries the last-pushed baseline onto the new branch so a later submit still has a baseline to check against.

Documented in the CR guide; added test scripts covering the advanced, diverged, and force-push cases.

### DIFF
--- a/.changes/unreleased/Fixed-20260620-071248.yaml
+++ b/.changes/unreleased/Fixed-20260620-071248.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'submit: Refuse to overwrite commits added to a branch by someone else since git-spice last pushed it; the offending commits are listed. Run ''gs branch sync'' to integrate them, or use --force to overwrite.'
+time: 2026-06-20T07:12:48.071369-04:00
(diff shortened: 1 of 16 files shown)


ed-irl:

- #1307
    - #1262
        - #1308 ◀

PR #1309, opened 2026-06-24
fix(repo sync): Handle reused manual branch names

abhinav:
`repo sync` looks up manually submitted changes by source branch name
before deciding whether a local branch is safe to delete.
That source branch name can be reused after an earlier manual change
has merged.

Before this patch,
the failure sequence was:

1. `repo sync` asked the forge for recent changes with the branch name.
2. A newer open change appeared before an older merged change.
3. `repo sync` selected the open change.
4. The older merged change was never checked against the local branch head.
5. The local branch stayed behind even though its commits were merged.

The regression script builds that boundary with two CRs that both use
`feature` as the source branch.
With the production selection logic reverted,
the script fails because `repo sync` never reports the older merged CR:

    no match for `feature: #1 was merged` found in stderr

`repo sync` now resolves the local branch head before choosing among
manual change results.
A merged change whose forge head contains the local head proves that
the local branch is safe to delete,
so that merged change outranks an open change returned for the same
reused source branch name.

The same script now verifies that `repo sync` reports the merged CR,
deletes the local `feature` branch,
and leaves the newer open CR intact.

### DIFF
--- a/.changes/unreleased/Fixed-20260622-182305.yaml
+++ b/.changes/unreleased/Fixed-20260622-182305.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'repo sync: Detect manually submitted merged changes when a source branch name is reused by a newer open change.'
+time: 2026-06-22T18:23:05.333324-07:00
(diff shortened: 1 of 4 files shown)


abhinav:

- #1309 ◀

PR #1313, opened 2026-06-26
fix(submit): Refuse trunk submit upstreams

### LINKED ISSUE #1311: Bug: `gs branch submit` pushes a tracked branch onto the trunk ref instead of refusing
**Tool:** git-spice (`gs`) 0.29.2
**Platform:** macOS (aarch64-darwin), forge: GitHub

## Summary

When a branch tracked by git-spice has its **git upstream** configured to the
trunk's remote ref (`origin/main`), `gs branch submit` resolves the push
destination from that upstream and **pushes the branch's commits onto the trunk
branch (`origin/main`)** rather than to `origin/<branch>` — or refusing. git-spice
already knows `main` is the trunk (it is the branch's configured base and the
repo trunk), so pushing a non-trunk branch's commits onto it is never the intent
of `branch submit`.

On the first submit, gs derives the upstream name from git's `branch.<name>.merge`
ref, then persists it as the branch's git-spice upstream in `refs/spice/data`. Once
cached, the push destination comes from gs's own state, so a later
`git branch --unset-upstream` no longer changes it.

It prints an informational notice ("Using upstream name 'main'") but proceeds to
push anyway, then fails only later at PR creation ("No commits between main and
main"), leaving the trunk advanced and no PR created.

## Reproduction

1. Create a branch in a worktree off the remote trunk ref (this sets the branch's
   upstream to `origin/main`):
   ```
   git worktree add -b feat/x /path/wt origin/main
   # git: "branch 'feat/x' set up to track 'origin/main' by rebasing"
   ```
2. Commit on `feat/x`.
3. Track it in git-spice with the trunk as base:
   ```
   gs branch track --base main      # "feat/x: tracking with base main"
   ```
4. Submit:
   ```
   gs branch submit --draft --fill
   ```

## Observed (actual output)

```
INF feat/x: Using upstream name 'main'
INF feat/x: If this is incorrect, cancel this operation and run 'git branch --unset-upstream feat/x'.
DBG Pushing to remote  name=origin lease.ref=main lease.hash=<trunk-sha>
DBG git push: To github.com:owner/repo.git
DBG git push:    <trunk-sha>..<feat-sha>  <feat-sha> -> main
DBG Setting branch upstream  name=feat/x upstream=origin/main
FTL gs: submit branch feat/x: publish change: create change: create pull request: Post "https://api.github.com/graphql": createPullRequest: UNPROCESSABLE: No commits between main and main
```

Net effect: `feat/x`'s commit was pushed onto `origin/main` (a lease-guarded
fast-forward of the trunk), the branch's upstream was set to `origin/main`, and
PR creation then failed because head and base resolved to the same branch.

## Root cause

The push destination is chosen in [`resolveUpstreamBranch`](https://github.com/abhinav/git-spice/blob/main/internal/handler/submit/handler.go)
(`internal/handler/submit/handler.go`). Its precedence is: (1) the upstream stored
in git-spice state, (2) the branch's git upstream if it points at the submit remote,
(3) none. Step 2 strips the `origin/` prefix off the git upstream and returns the
remainder as the push target — with no check against the trunk:

```go
if b, ok := strings.CutPrefix(upstream, remote+"/"); ok {
    h.Log.Infof("%v: Using upstream name '%v'", branch, b)
    return b, nil
}
```

For an upstream of `origin/main`, `b` is `main`, the trunk. `submitBranch` already
guards `branchToSubmit == Store.Trunk()` ("cannot submit trunk") and has
`Store.Trunk()` in scope, so the trunk name is known at the exact point the
destination is resolved — the existing guard just doesn't cover pushing a non-trunk
branch *onto* the trunk ref.

## Expected

`gs branch submit` should not push a non-trunk branch's commits onto the trunk.
When the resolved push destination equals the repo's configured trunk branch, it
should **abort with a clear error** (e.g. "refusing to push branch 'feat/x' to
trunk 'main'; its upstream is the trunk — run `git branch --unset-upstream`") and
push nothing. If a same-name fallback is ever wanted, it should be opt-in behind
an explicit flag, not the silent default.

Suggested fix location: `resolveUpstreamBranch` in
`internal/handler/submit/handler.go` — when the resolved upstream name (precedence
step 2) equals `Store.Trunk()`, abort instead of returning it.

The current "Using upstream name 'main'" notice is informational only and is
printed inline as the push runs (it is not an interactive prompt; gs does not wait
for acknowledgment before pushing), so there is no actual opportunity to cancel
before the trunk is modified.

## Impact

- Silently advances the trunk with un-reviewed commits, bypassing the PR flow
  `submit` exists to create.
- Leaves the repo in a half-done state (trunk moved, no PR).
- Remediation requires force-pushing the trunk back, which is dangerous and is
  frequently blocked by branch protection or by agent/safety guards.
- This is a regression from stock git safety: plain `git push` under the default
  `push.default=simple` already refuses a branch whose upstream name differs from
  the branch name, so gs pushing anyway is not user error.

## Workarounds

- Before the *first* `submit`, ensure the branch's upstream is not the trunk:
  `git branch --unset-upstream <branch>` (or set it to `origin/<branch>`). After a
  submit has already cached the gs-upstream, unsetting git's upstream is not enough;
  reset the cached git-spice upstream too (re-track via `gs branch untrack` then
  `gs branch track --base main`).
- Create branches with `gs branch create` rather than off `origin/main`, so the
  upstream is never the trunk ref.

## Notes

This was hit in real use: a worktree branch created with
`git worktree add -b feat/x <path> origin/main` carried `origin/main` as its
upstream; `gs branch track --base main` (correct, standard invocation) set the
git-spice base; then `gs branch submit --draft --fill` pushed the commit to
`origin/main`. The `branch track` step did not cause the push — the push target
came from the git upstream — but git-spice had `main` registered as trunk both
ways (base and trunk) and still pushed to it.

abhinav:
branch submit uses an upstream branch name from git-spice state or from
the branch's Git upstream to choose the remote head it pushes.

If that upstream resolves to trunk, a non-trunk branch can push its commits
onto the trunk ref before change request creation fails. Refuse both stored
and Git-derived trunk upstreams before building the push refspec, with
recovery guidance for the state that supplied the unsafe upstream.

The script regression proves the fresh Git upstream case by showing
origin/main remains unchanged and no change request is created. The unit
test covers cached git-spice upstream state equal to trunk.

Resolves #1311

### DIFF
--- a/.changes/unreleased/Fixed-20260625-225551.yaml
+++ b/.changes/unreleased/Fixed-20260625-225551.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: "branch submit: Accidental pushes to trunk are now prevented when a branch's upstream is trunk."
+time: 2026-06-25T22:55:51.101877222Z
--- a/internal/handler/submit/handler.go
+++ b/internal/handler/submit/handler.go
@@ -1226,6 +1226,13 @@ func (h *Handler) resolveUpstreamBranch(
 	// Stored upstream branch state wins because it reflects the branch name
 	// previously used for submission, even if the local branch was renamed.
 	if storedUpstream != "" {
+		if storedUpstream == h.Store.Trunk() {
+			return "", fmt.Errorf(
+				"refusing to push branch %q to trunk %q; "+
+					"run '%s branch untrack %s' and track the branch again",
+				branch, storedUpstream, cli.Name(), branch)
+		}
+
 		return storedUpstream, nil
 	}
 
@@ -1241,6 +1248,13 @@ func (h *Handler) resolveUpstreamBranch(
 	}
 
 	if b, ok := strings.CutPrefix(upstream, remote+"/"); ok {
+		if b == h.Store.Trunk() {
+			return "", fmt.Errorf(
+				"refusing to push branch %q to trunk %q; "+
+					"run 'git branch --unset-upstream %s'",
+				branch, b, branch)
+		}
+
 		h.Log.Infof("%v: Using upstream name '%v'", branch, b)
 		h.Log.Infof("%v: If this is incorrect, cancel this operation and run 'git branch --unset-upstream %v'.", branch, branch)
 		return b, nil
--- a/internal/handler/submit/handler_test.go
+++ b/internal/handler/submit/handler_test.go
@@ -185,6 +185,52 @@ func TestHandler_SubmitBatch_rejectsStaleBaseBeforeSubmitting(t *testing.T) {
 	assert.Contains(t, logBuffer.String(), "base=feat1")
 }
 
+func TestHandler_resolveUpstreamBranch_refusesStoredTrunk(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+
+	mockStore := NewMockStore(mockCtrl)
+	mockStore.EXPECT().
+		Trunk().
+		Return("main")
+
+	handler := &Handler{
+		Log:        silog.Nop(),
+		View:       ui.NewFileView(&bytes.Buffer{}),
+		Repository: nil,
+		Worktree:   nil,
+		Store:      mockStore,
+		Service:    nil,
+		Browser:    &browser.Noop{},
+		FindRemote: func(context.Context) (state.Remote, error) {
+			return state.Remote{}, nil
+		},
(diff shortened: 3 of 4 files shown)


abhinav:

- #1313 ◀

PR #1314, opened 2026-06-27
fix(git): Preserve merge-tree wait errors

abhinav:
Repository.MergeTree parses git merge-tree stdout before it waits for the process to finish.
If parsing succeeded but cmd.Wait failed,
the old error path wrapped the stale parse error variable instead of the wait error.
That produced git merge-tree: %!w(<nil>) and hid the real process failure.

Wrap waitErr so callers receive the command-completion failure and can still unwrap it with errors.Is.
The regression test fails against the old implementation with the malformed nil wrap,
and passes after the wait error is preserved.

[skip changelog]: internal diagnostic correctness fix


abhinav:

- #1314 ◀

PR #1315, opened 2026-06-28
mergequeue: Schedule dependent merges in phases

abhinav:
Divergent stack merges cannot be handled as a linear queue.

Before this change,
the scheduler effectively kept one branch in the active lane.
For a stack shaped like this:

```
    main
      |
    feat1
     / \
 feat2 feat3
```

the merge loop could merge feat1,
then prepare and wait on feat2 before feat3 was even restacked:

    time ->
    feat1: | prep | run | sync |
    feat2:                     | prep | run | sync |
    feat3:                                         | prep | run | sync |

That left independent branches idle even though feat2 and feat3 only
depended on feat1, not on each other.
This would be made worse if feat2 was blocked on merge readiness:

    feat2: | prep | run..... | sync |
    feat3:                          | prep | run | sync |

This change reshapes mergequeue around graph items with one parent.
An item with an empty parent is a root.
When an item's parent is done,
the scheduler can prepare that item.

Item.Prepare is serialized by the scheduler.
This preserves the caller's ability to do local preparation work
without adding locks around each item implementation.

Item.Run may execute concurrently with other ready items.
A Run call reports the asynchronous operation for that item:
for merge handling,
that means waiting for forge readiness,
requesting the merge,
and waiting for the forge to report the change merged.

Options.Barrier is a queue-level synchronization point.
A successful Item.Run marks the barrier dirty.
Before the scheduler starts later prepares,
it runs the barrier once and coalesces any sibling runs
that completed before the barrier started.
The scheduler also attempts a dirty barrier before returning,
including after the final successful item.

The merge handler uses the barrier for SyncTrunk.
That lets sibling merge requests proceed independently,
while still guaranteeing that local trunk synchronization happens
before preparing branches stacked above merged branches.

With the same divergent stack,
the scheduler can now prepare and run independent branches in phases,
which means that for the two divergent branches:

    time ->
    feat1: | prep | run | sync |                   |      |
    feat2:              |      | prep | run..... ||| sync |
    feat3:              |      |      | prep | run |      |

Besides allowing divergent branches to merge independently,
this has the advantage of reducing the number of syncs
by coalescing merges that complete in the same phase
into one SyncTrunk call.

### DIFF
--- a/branch_merge.go
+++ b/branch_merge.go
@@ -25,9 +25,9 @@ func (*branchMergeCmd) Help() string {
 		The branch must be based directly on trunk.
 		To merge a stacked branch, use 'gs downstack merge'.
 
-		Before merging, waits for merge readiness:
-		the forge must observe the pushed head
-		and report that the CR is ready to merge.
+		Before checking merge readiness,
+		the command waits briefly for the forge to observe the pushed head.
+		Then it waits for the forge to report that the CR is ready to merge.
 		Use --ready-timeout to configure the maximum wait.
 	`)
 }
--- a/doc/includes/cli-reference.md
+++ b/doc/includes/cli-reference.md
@@ -310,9 +310,9 @@ Before merging, the stack is checked for branches
 whose base PR was already merged on the forge.
 Use --no-branch-check to skip this validation.
 
-Before each merge, waits for merge readiness:
-the forge must observe the pushed head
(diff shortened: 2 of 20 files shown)


abhinav:

- #1315 ◀
    - #1318
    - #1316

PR #1316, opened 2026-06-28
mergeprogress: Add glyph set presets

abhinav:
Replace merge progress bar glyphs with a better set.
Glyphs all have the same height now for a more balanced bar shape.

demo

ASCII glyph set is available for a future configuration option.


abhinav:

- #1315
    - #1316 ◀

PR #1317, opened 2026-06-28
fix(shamhub): Stream template lookup paths

abhinav:
Submitting a stack with no change templates could time out while ShamHub
walked each possible template path through failing Git object lookups.
The failing script printed this warning before each CR:

    WRN Could not list change templates
    error=list templates: lookup change body template:
    send HTTP request:
    Get ".../change-template": context deadline exceeded

Missing templates are an expected state,
so ShamHub should not surface template lookup warnings for that case.

Stream the repository tree with the existing NUL scanner
and match only candidate template files before reading blobs.
This keeps the no-template path quiet while preserving case-insensitive file
and directory-template matching.

The upstack_submit_main script failed before this patch
because the warning lines polluted the expected submit log,
and passes after the change.

### DIFF
--- a/internal/forge/shamhub/template.go
+++ b/internal/forge/shamhub/template.go
@@ -3,10 +3,12 @@ package shamhub
 import (
 	"context"
 	"fmt"
+	"path"
 	"slices"
 	"strings"
 
 	"go.abhg.dev/gs/internal/forge"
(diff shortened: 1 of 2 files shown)


abhinav:

- #1317 ◀

PR #1318, opened 2026-06-28
fix(shamhub): Use singular change URLs

abhinav:
ShamHub REST use the singular `/change/<num>` path
but `RepositoryID.ChangeURL` was using `/changes/<num>`.

Make them use the same path and update fixtures.


abhinav:

- #1315
    - #1318 ◀

PR #1323, opened 2026-06-29
merge: Add merge command hook

abhinav:
Repositories can configure `spice.merge.command` to request a merge through a user-defined command instead of the forge merge API. The command runs after git-spice observes the pushed head and after the forge reports that the change is ready to merge.

The hook environment includes common `GIT_SPICE_*` branch, forge, URL, trunk, base, and head-SHA variables. Forge-specific identifiers use provider-specific names, and forge-provided values cannot overwrite common variables.

The command is only the merge request step. A successful command still flows into the existing await-merged path, and a non-zero command exit fails that merge queue item. Command output streams through the merge logger as it is produced instead of being buffered into the script result.

Validation covers the handler path, the script runner streaming contract, successful merge command execution through ShamHub, and merge command failure without a forge merge request.

[skip changelog]: merge command support is unreleased.

### DIFF
--- a/doc/includes/cli-reference.md
+++ b/doc/includes/cli-reference.md
@@ -331,7 +331,7 @@ Use --fail-fast to stop the queue after the first branch failure.
 * `--fail-fast`: Stop the merge queue after the first branch failure.
 * `--branch=NAME,...`: Branches whose stacks to merge. May be repeated.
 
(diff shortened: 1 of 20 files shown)


abhinav:

- #1323 ◀

PR #1325, opened 2026-06-29
merge: Order progress slots by dependency layer

abhinav:
The merge progress bar is a flat horizontal display,
and branch names are intentionally kept out of the bar.
Previously, order of items in the bar was unspecified.

This changes it to use a BFS ordering.
For example, given these merge dependencies:

```text
trunk
├─ feat1
│  ├─ feat3
│  │  └─ feat6
│  └─ feat4
├─ feat2
│  └─ feat5
└─ feat7
```

The progress bar now receives this order:

```text
feat1 feat2 feat7 feat3 feat4 feat5 feat6
```

The effect is that roots are displayed first,
then their immediate children (in the same order as the roots),
then deeper descendants.
While progress won't move left-to-right all the time,
this is closer to that expectation.

demo

### DIFF
--- a/internal/graph/bfs.go
+++ b/internal/graph/bfs.go
@@ -1,0 +1,63 @@
+package graph
+
+import "go.abhg.dev/gs/internal/must"
+
+// BreadthFirstSort orders nodes by breadth-first dependency layers.
+//
+// parent returns the index of a node's parent in nodes,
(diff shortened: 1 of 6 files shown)


abhinav:

- #1325 ◀

PR #1330, opened 2026-07-02
fix(submit): Avoid terminal hang after template warning

### LINKED ISSUE #1329: branch submit hangs and then breaks terminal display
Since updating to 0.30.0 `branch submit` hangs for me after entering the title (the font goes to slightly darker after pressing enter and then nothing happens anymore).

After then eventually killing the process, my terminal (zsh) gets broken somehow. Might be an issue with my setup?

Image

Setting GIT_SPICE_VERBOSE to true only gives my repo access level as additional output.

abhinav:
## Problem

branch submit could hang after the title prompt
if CR template lookup returned a warning
while the interactive form was waiting for the template result.

```
script error: await: "Body:" not found
signal: killed
```

The warning path wrote through OutputWriter
while Bubble Tea owned the terminal.
OutputWriter routed that write through Program.Send,
but the Bubble Tea update loop was waiting
for the same background result that produced the warning.

## Solution

Route completed output lines through a background goroutine.

Write now only buffers completed-line input and wakes the goroutine,
so warning output cannot block the caller waiting for the form.

OutputWriter.Close stops the goroutine to avoid leakage.
This was not incorporated into main()
as that would require a more invasive refactor in exchange for no value
(as the program is exiting anyway).

## Testing

The test script delays ShamHub template lookup and causes the warning.
A similar failure was reproduced in the test script without the fix.

The output writer unit test also verifies this issue
by blocking the Bubble Tea update loop and verifying that Write returns.

---

Resolves #1329

### DIFF
--- a/.changes/unreleased/Fixed-20260702-053109.yaml
+++ b/.changes/unreleased/Fixed-20260702-053109.yaml
@@ -1,0 +1,5 @@
+kind: Fixed
+body: >-
+  branch submit: Fix deadlock and terminal display corruption
+  if CR template lookup failed during interactive prompt.
+time: 2026-07-02T05:31:09.777099-07:00
--- a/internal/forge/shamhub/cli.go
+++ b/internal/forge/shamhub/cli.go
@@ -260,6 +260,16 @@ runCommand:
 
 		key, value := args[0], args[1]
 		switch key {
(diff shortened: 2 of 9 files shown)


abhinav:

- #1330 ◀

PR #1332, opened 2026-07-02
refactor(handler/restack): Move options into requests

abhinav:
Extracted from #1307

Restack handler entry points now receive operation request structs
instead of loose branch and options parameters. The request structs carry
optional Options fields, matching the handler request pattern used by the
surrounding handler packages.

UpstackOptions embeds the shared Options type while keeping SkipStart as
the only active option. AutoResolve is intentionally not part of this
commit.


abhinav:

- #1332 ◀

PR #1336, opened 2026-07-04
test(merge): Stabilize stack fail-fast test

abhinav:
The stack merge fail-fast test was asserting that every ready sibling enters preparation after the first sibling reports a blocked merge.

The scheduler is allowed to stop scheduling as soon as fail-fast observes that blocked result.
It may also have already handed another sibling to preparation by then.
Treat the sibling preparation call as optional so the test covers the fail-fast contract instead of one scheduler interleaving.

[skip changelog]: Test-only change.


abhinav:

- #1336 ◀

PR #1339, opened 2026-07-05
submit: Move submit option types into options file

abhinav:
Keep the submit handler workflow in handler.go while moving the command and configuration option declarations into options.go. The option modes and merge helpers stay with Options so future option changes can be reviewed without crossing the submit publication workflow.

This is a package-local organization change. It does not change submit command flags, configuration keys, exported handler APIs, or publication behavior.

[skip changelog]


abhinav:

- #1339 ◀

PR #1343, opened 2026-07-05
branchtree: Model highlighted renderer text locally

abhinav:
Introduce a package-local highlightedText value for renderer text paired with highlight indexes. Renderer helpers now pass that value through branch names, change IDs, worktree paths, and final text rendering instead of carrying loose string and []int pairs.

The public Item fields stay unchanged, so callers can continue to populate Branch, ChangeID, Worktree, and their highlight indexes directly. Branchtree tests now parse marked-up test strings into the same local renderer value used by production code.

[skip changelog]


abhinav:

- #1343 ◀

PR #1346, opened 2026-07-05
fix(uitest): Classify each RunScripts input

abhinav:
RunScripts promises that each input path can be a script file or a directory containing script files. The input scan now stats the current path instead of reusing the first path, so mixed file and directory inputs both enter testscript with the intended script files.

The regression covers a file argument followed by a directory argument and fails when the directory is classified from the first file. The nearby SetupScript and ScriptOptions documentation now matches the script API.

[skip changelog]


abhinav:

- #1346 ◀

PR #1351, opened 2026-07-07
test(mergequeue): Stabilize barrier coalescing test

abhinav:
The barrier coalescing test needs `block-prepare` to be inside
`Prepare` before it releases the sibling `Run` calls.
Without that interlock, CI scheduling can let the barrier observe only
one completed sibling before the blocking prepare phase starts.

Wait for the blocking `Prepare` call to start before releasing sibling
runs.
The test now covers the intended scheduler state transition instead of
relying on goroutine timing.


abhinav:

- #1351 ◀

PR #1365, opened 2026-07-12
github: Batch identity lookups

abhinav:
Resolve uncached reviewer and assignee identities with one aliased
GraphQL query per metadata operation. Cache successful user and team
lookups while preserving ordered results and missing-identity errors.

Single-identity requests use the same batching path. Recorded reviewer
and assignee fixtures cover the new wire representation.

PR #1364, opened 2026-07-12
github: Batch label ID lookups

abhinav:
Resolve requested label names with one aliased GraphQL query before creating missing labels. Preserve input order, suppress duplicate lookups and creations, and create distinct missing labels concurrently.

Keep the direct single-label path for create-race recovery.

Validation covers deterministic query construction, variable escaping, missing labels, duplicate association, concurrency, and request counts. The GitHub integration fixture records the batched request.

PR #1363, opened 2026-07-12
github: Hide GraphQL pagination in gateway

abhinav:
Move comment and status-check traversal behind gateway iterators, and make batched review-thread counts traverse continuation pages internally. Forge callers no longer own cursors or continuation loops.

Preserve the existing GraphQL documents, variables, default page sizes, and request order. `PaginationOptions` permits smaller pages for callers and tests without exposing cursors.

Validation strengthens gateway coverage for multi-page request documents, cursor variables, result order, cardinality, and continuation errors. The existing GitHub forge replay passes without fixture changes.

PR #1362, opened 2026-07-12
github: Add GraphQL gateway

abhinav:
Replace shurcooL/graphql with custom GitHub GraphQL gateway/client
with domain-scoped operations implemented locally.

Immediately, this allows deleting the weird graphqlutil error hack.

Soon after, this will allow use of dynamic GraphQL queries
which were not possible to represent with reflection-based
query specification in shurcooL/graphql,
allowing batching of API calls that are currently done separately.

PR #1366, opened 2026-07-13
github: Batch PR metadata updates

abhinav:
Resolve label and identity inputs before sending pull request labels,
review requests, and assignments as one aliased GraphQL mutation.
Resolve label and identity lookups concurrently, and share one identity
lookup across reviewers and assignees.

Keep base and draft transitions separate because GitHub may execute later
nullable mutation fields after an earlier field fails. Preserve stable
aliases in GraphQL error paths and do not retry partial results.

Enable a shared integration scenario that submits one change with both a
reviewer and an assignee. Record the scenario against GitHub, Gitea,
Forgejo, and ShamHub; GitLab and Bitbucket remain explicitly skipped with
follow-up TODOs.

## Validation

The full repository test suite passes with the refreshed GitHub fixtures
and the locally recorded Gitea, Forgejo, and ShamHub fixtures.


abhinav:

- #1362
    - #1363
        - #1364
            - #1365
                - #1366 ◀

PR #1368, opened 2026-07-13
release/ppa: allow opting out of orig.tar.gz upload

abhinav:
Re-publish attempts after initial upload should not include orig.tar.gz
because that's already published in the first attempt.
Add an option to opt out of it.


abhinav:

- #1368 ◀
    - #1369

PR #1372, opened 2026-07-18
fix: Exclude stale base commits from replays

abhinav:
Branch restack and branch onto now replay only commits owned by the branch when a tracked branch was rebased outside git-spice onto a newer base.

An external rebase can leave the recorded base as an older ancestor of the branch. If the base advances again, replaying from that recorded hash includes base-branch commits and can conflict before reaching the branch commits.

Resolve each replay boundary from the recorded base, current merge base, and branch head. Advance stale ancestral boundaries, preserve newer boundaries from retarget-without-rebase, and use fork-point recovery only when the recorded boundary is disconnected. Restack checks, restack execution, and `BranchOnto` share this decision.

Stack edit coverage exercises the same behavior through `BranchOnto`. It is evidence for that consumer path, not a separately changed replay surface.

## Validation

- The branch-restack regression externally rebases a two-commit branch, advances its base with a conflicting edit, and protects both commits.
- The branch-onto regression moves the same stale-boundary shape to an independent destination and rejects replay of the old base commit.
- The stack-edit regression reorders a stack through `BranchOnto` and protects only the feature commits and destination content.
- Compatibility cases protect ordinary base advances from a poisoned fork point and retain newer retarget-without-rebase boundaries.

### DIFF
--- a/.changes/unreleased/Fixed-20260718-090808.yaml
+++ b/.changes/unreleased/Fixed-20260718-090808.yaml
@@ -1,0 +1,5 @@
+kind: Fixed
+body: >-
+  branch restack, branch onto: Conflicting rebases caused by replayed base
+  commits are avoided after a branch is rebased outside git-spice.
+time: 2026-07-18T09:08:08.569856-07:00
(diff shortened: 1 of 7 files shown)


abhinav:

- #1372 ◀

PR #1393, opened 2026-08-06
Pad leaf connectors in tree view to fix sibling branch label alignment

### LINKED ISSUE #1383: Leaf branches misalign with sibling branches that have an upstack in list command
When two branches share the same parent, but one has an upstack and the other doesn't, their labels don't line up. The branch with children gets a `┻` junction glyph, which shifts its node marker and label one column to the right compared to the leaf sibling.

Example (`gs lsa`):

```
┃     ┏━┻□ branch-a
┃     ┣━□ branch-b
┃   ┏━┻■ branch-c ◀
```

`branch-a` (`┏━┻□`) is 4 glyphs wide, `branch-b` (`┣━□`) is 3, so the two siblings' `□` markers and labels are offset by one column.

Padding the leaf connector with one extra `━` would align them:

```
┃     ┏━┻□ branch-a
┃     ┣━━□ branch-b
┃   ┏━┻■ branch-c ◀
```

Minor cosmetic thing, but same-level branches reading at slightly different indents is a little confusing at a glance.

Related: https://github.com/abhinav/git-spice/blob/c2a06eb5b890f261f86974bdd51bcb8f1f8ae90f/internal/ui/fliptree/tree.go#L189

Armaldio:
Fix https://github.com/abhinav/git-spice/issues/1383

When two branches share the same parent but one has children and the other doesn't, their labels don't line up in gs ls and related commands. The branch with children gets a ┻ junction glyph, which shifts its node marker and label one column to the right compared to the leaf sibling.

Before:
```
┃     ┏━┻□ branch-a
┃     ┣━□ branch-b
```
After:
```
┃     ┏━┻□ branch-a
┃     ┣━━□ branch-b
```

The fix adds a horizontal segment to leaf node connectors so they match the width of the junction glyph. This keeps same-level branches reading at the same indent.

Added a regression test covering the leaf/parent sibling case
Updated all affected fixtures (unit tests, widget scripts, and integration scripts)
Added changelog entry

### DIFF
--- a/.changes/unreleased/Fixed-20260805-113924.yaml
+++ b/.changes/unreleased/Fixed-20260805-113924.yaml
@@ -1,3 +1,5 @@
 kind: Fixed
-body: 'tree: Pad leaf node connectors to match width of junction glyphs, aligning sibling branch labels.'
+body: >-
+  Sibling branches in rendered branch trees are no longer unaligned
+  when one of them has upstack branches and the other does not.
 time: 2026-08-05T11:39:24+00:00
--- a/.changes/unreleased/Fixed-20260805-113924.yaml
+++ b/.changes/unreleased/Fixed-20260805-113924.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: 'tree: Pad leaf node connectors to match width of junction glyphs, aligning sibling branch labels.'
+time: 2026-08-05T11:39:24+00:00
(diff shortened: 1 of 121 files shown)


Armaldio:
Not sur everything is correct, please have a look :pray:

abhinav:
Hey, will look at this as soon as I'm able. The change should be fine.

abhinav:
This is great. Thank you.

PR #1396, opened 2026-08-10
github: Add native stack gateway operations

abhinav:
GitHub's native stack and asynchronous merge APIs require REST
alongside the existing GraphQL operations. Keep both protocols behind
one Gateway so callers use repository-domain operations instead of
transport clients.

Expose only the stack mutations and asynchronous merge state needed by
the forge layer. Attach to existing pending merges by operation ID
without changing merge behavior.

Stack reconciliation needs only pull request state, head repository
identity, and native stack membership. Add a compact batch lookup that
preserves input order, leaves missing pull requests nil, and loads each
distinct native stack once instead of loading full pull request
metadata.

### DIFF
--- a/internal/gateway/github/async_merge.go
+++ b/internal/gateway/github/async_merge.go
@@ -1,0 +1,193 @@
+package github
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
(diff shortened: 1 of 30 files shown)


abhinav:

- #1396 ◀
    - #1397
        - #1398
            - #1399
                - #1405
                    - #1400
                        - #1401
                            - #1403
                                - #1406

PR #1404, opened 2026-08-12
fix(race): Prioritize merge queue cancellation

abhinav:
Cancellation-aware workers can return at the same time that the caller
context becomes done. The selected case previously determined whether
pending children were skipped as canceled or below a failed parent.

Check the caller context before processing a selected worker result.
Apply shutdown semantics to that result after canceling pending work.
This preserves started-work failures while giving pending items a stable
cancellation reason.


abhinav:

- #1404 ◀

Issue #1388, opened 2026-07-30
Native GitHub stacks support


abhinav:
https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests is public. 
There are APIs that we can take advantage of.

From a quick read: stacks must be "created", and PRs must be added to them.
Stacks are only allowed to be linear, i.e. A -> B -> C, and A -> B -> D cannot coexist.

Immediate thoughts:

- `submit` operations will want to auto-manage stack resources. When a stack is linear, auto-add to stack.
    When a stack becomes non-linear, choose the longest path to keep stacked per GitHub,
    If there is ambiguity (multiple equally long paths), tie-break using some deterministic criteria (earliest PR number?).
- at `merge` time, if a run of branches is part of the same stack, merge them atomically with GitHub
- `merge` may want to unconditionally switch to GitHub's async merge API going forward

abhinav:
Current status: almost there.

- #1396
- #1397
- #1398
- #1399
- #1400
- #1401
- #1403

Last two PRs need some cleaning up, docs after that and this should be good to go.

PR #1403, opened 2026-08-11
submit: Finalize partial submissions

abhinav:
A batch submit returned immediately when any branch failed. Earlier change
requests had already been published, but their native stack membership and
navigation comments were skipped until another successful submit.

Defer stack-representation synchronization for completed branches and join its
error with the primary submit failure. Load one post-submit branch graph, derive
the native stack request before navigation metadata can change, and perform the
independent forge updates concurrently. Native stack failures remain
warning-only, while navigation comment failures and the original submit error
stay observable.

PR #1401, opened 2026-08-10
merge: Use native ranges when available

abhinav:
Merge selected linear paths atomically when a repository implements
`WithMergeRange`. Prepare each change and verify its published base,
head, and commit before scheduling the operation, then use the existing
timeout for asynchronous acceptance and final change-state polling.

Treat runtime `ErrUnsupported` as a signal to resume the ordinary
bottom-up merge sequence. Divergent heads remain separate queue
dependencies, so one native linear path can merge without consuming an
unselected sibling.

ShamHub constructs merge and squash range results before one
compare-and-swap target update. Propagate request cancellation through
validation, commit construction, and publication so a canceled operation
cannot advance the target or mark changes merged.

A GitHub-only integration scenario registers A-B-C as a native stack,
merges it atomically, and verifies that divergent D remains open,
unstacked, and mergeable on B with its head branch and commit unchanged.
Its cassette remains deliberately unrecorded for a human operator to
capture.

PR #1400, opened 2026-08-10
submit: Update native stacks after publishing

abhinav:
Submission can create or update change requests without refreshing a
forge's native stack metadata. After successful publication, reload the
branch graph and send each affected tracked tree to repositories that
implement `WithStacks`.

Use `BranchGraph.Bottom` to find each affected tree and collect its
published changes. Treat an unsupported capability as a no-op, and warn
for other update failures without changing the successful submission
result.

Model the capability in ShamHub behind an opt-in configuration so
end-to-end tests can observe complete, order-independent native stack
trees without changing existing script behavior.

PR #1405, opened 2026-08-15
shamhub: Model native stack operations

abhinav:
Submit and merge integration need a deterministic forge that exposes the same
optional native-stack boundary as production providers.
Add an opt-in ShamHub repository implementing the complete `WithStacks`
contract while leaving the capability absent by default for existing scripts.

Plan immutable stack replacements without mutating stored state.
Execution atomically validates and replaces the touched relationships together
with their provider-facing bases, allowing submit tests to detect sequencing
regressions.

Preserve divergent relationships in storage and derive disjoint linear merge
plans from that representation.
Each executable range validates and constructs its merge or squash result
before one compare-and-swap publication, so failures cannot expose a partial
atomic merge.

Refs #1388

### DIFF
--- a/internal/forge/shamhub/admin.go
+++ b/internal/forge/shamhub/admin.go
@@ -1,6 +1,7 @@
 package shamhub
 
 import (
+	"cmp"
 	"context"
 	json "encoding/json/v2"
 	"errors"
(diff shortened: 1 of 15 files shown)


abhinav:

- #1396
    - #1397
        - #1398
            - #1399
                - #1405 ◀
                    - #1400
                        - #1401
                            - #1403
                                - #1406

PR #1419, opened 2026-08-23
forge: Expose review thread commit hash

abhinav:
`ReviewThread` now reports the change head revision
against which its root comment was created.
Provider adapters use the forge-owned reviewed revision,
while unsupported products leave the hash zero.
Replies retain the root revision even when the change head advances.

Gitea and Forgejo take the revision from the containing review because
ordinary review comments leave `original_commit_id` empty.
Gitea associates each comment with its review before sorting synthetic
cross-review threads, so a later reply cannot replace the root revision.

ShamHub captures the local change head when root creation begins,
persists that revision on roots and replies through both submission paths,
and resolves local refs without holding its storage lock.

Co-Authored-By: Edmund Kohlwey <ed@irl.llc>


abhinav:

- #1419 ◀

PR #1428, opened 2026-08-30
submit: Resolve abbreviated change heads

### LINKED ISSUE #1421: bitbucket: HeadHash uses truncated commit hash from API, causing false "branch already in use" on submit
`branch submit` (and `stack submit`) never recognizes a pre-existing Bitbucket pull request for a branch that wasn't created by git-spice — it treats the branch name as "already in use with a different commit" and creates a new PR on a renamed branch (`<branch>-2`) instead of updating the existing one, even when the branch's HEAD matches the PR's source commit exactly.

## Root cause

`internal/gateway/bitbucket/cloud/change.go`:

```go
var headHash git.Hash
if pr.Source.Commit != nil {
    headHash = git.Hash(pr.Source.Commit.Hash)
}
```

This assigns `HeadHash` directly from Bitbucket Cloud's REST API field `source.commit.hash`. Bitbucket Cloud's API returns this field as a 12-character abbreviated hash, not the full 40-character SHA (unlike GitHub/GitLab, which return full SHAs). Nothing downstream expands/normalizes this short hash before comparing it against the full local branch HEAD, so the comparison always fails for any PR that predates git-spice tracking the branch.

## Steps to reproduce

1. Have an existing open Bitbucket PR for branch `foo`, created outside git-spice (e.g. via the web UI or another tool), where the branch's current HEAD matches the PR's source commit.
2. `git spice repo init`, then `git spice branch track` on `foo`.
3. `git spice branch submit --dry-run` on `foo`.

## Observed

```
INF foo: Ignoring CR #N with the same branch name: remote HEAD (<12-char-hash>) does not match local HEAD (<40-char-hash>)
INF foo: Branch name already in use in remote 'origin'
INF foo: Using upstream name 'foo-2' instead
INF WOULD create a CR for foo
```

Both hashes printed are in fact the same commit — the local hash is simply the correct full-length version of the short hash from the API (verified via `git rev-parse` on both the local branch and `origin/foo` — identical, matching what the Bitbucket UI shows for the PR's source commit).

## Expected

git-spice should recognize the existing PR as up to date (or needing an update) instead of trying to create a duplicate PR — e.g. by comparing only the shared prefix length, or by resolving the short hash against the actual branch ref before comparing.

## Environment

- git-spice 0.31.2 (linux/amd64)
- Bitbucket Cloud

abhinav:
Bitbucket Cloud reports pull request heads as abbreviated hashes.
Exact comparisons therefore caused `branch submit` to ignore matching
externally-created pull requests, create a duplicate on a suffixed branch,
and push branches that were already up to date.

Resolve forge-reported heads through the local repository before comparing
them with the full local commit.
Unknown or ambiguous abbreviations remain conservative mismatches.

Fixes #1421

### DIFF
--- a/.changes/unreleased/Fixed-20260830-124226.yaml
+++ b/.changes/unreleased/Fixed-20260830-124226.yaml
@@ -1,0 +1,6 @@
+kind: Fixed
+body: >-
+  Bitbucket Cloud: `branch submit` now reuses an existing pull request for an
+  unchanged branch instead of creating a duplicate pull request on a suffixed
+  branch or pushing the branch again.
+time: 2026-08-30T12:42:26.103623-07:00
--- a/internal/handler/submit/handler.go
+++ b/internal/handler/submit/handler.go
@@ -454,7 +454,7 @@ func (h *Handler) submitBranch(
 			// If not, pretend we didn't find a matching CR.
 			if upstreamBranch == "" {
 				change := changes[0]
-				if change.HeadHash != commitHash {
+				if !h.changeHeadMatchesCommit(ctx, change.HeadHash, commitHash) {
 					log.Infof("%v: Ignoring CR %v with the same branch name: remote HEAD (%v) does not match local HEAD (%v)",
 						branchToSubmit, change.ID, change.HeadHash, commitHash)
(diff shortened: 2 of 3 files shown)


abhinav:

- #1428 ◀

PR #1429, opened 2026-08-30
repo sync: Detach merged branch worktrees

### LINKED ISSUE #1424: repo sync: auto-detach other worktrees
repo sync refuses to delete branches that are checked out in other worktrees even if they're merged.
we should have an opt-in (or out?) configuration that will instead auto-detach those worktrees rather than delay cleanup.

abhinav:
`spice.repoSync.detachWorktrees` lets `repo sync` detach another
worktree at its current commit before deleting its merged branch.
Staged, unstaged, and untracked changes remain in the detached worktree.

The setting defaults to false and preserves the existing warning-and-skip
behavior.
If the worktree cannot be opened, detached, or inspected after detaching,
`repo sync` warns once and leaves the branch intact.
Independent deletion candidates continue to be processed.

Fixes #1424


abhinav:

- #1429 ◀

PR #1434, opened 2026-09-05
jsonmut: Compose replayable transformations

abhinav:
Optimistic callers may apply the same transformation more than once as
concurrent writes replace the source document.

Provide immutable statements and typed programs for reading and rewriting
object members. Programs can sequence independent mutations or derive later
operations and results from earlier values without a mutable result registry.
Stream object rewrites so unrelated JSON does not require full decoding.


abhinav:

- #1434 ◀
    - #1435
        - #1436

PR #1435, opened 2026-09-05
storage: Add conflict-safe JSON updates

abhinav:
A caller that reads, transforms, and writes one document can overwrite a
concurrent update derived from a newer store revision.
Rebase continuation updates had that shape: concurrent appends could lose
work, and concurrent takes could return the same continuation more than once.

Expose pinned snapshots and compare-and-swap commits across each backend.
Apply replayable `jsonmut` programs to one snapshot and retry conflicts
against current state.
Move continuation appends and takes onto this path so each queue operation is
linearizable.

Required keys are checked in the same revision.
A take from a missing queue remains a no-op, and ordinary updates retain their
existing retry and idempotent no-op behavior.

### DIFF
--- a/internal/spice/state/continue.go
+++ b/internal/spice/state/continue.go
@@ -2,23 +2,21 @@ package state
 
 import (
 	"context"
+	"encoding/json/jsontext"
+	json "encoding/json/v2"
 	"errors"
 	"fmt"
+	"slices"
+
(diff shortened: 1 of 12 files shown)


abhinav:

- #1434
    - #1435 ◀
        - #1436

PR #1436, opened 2026-09-05
storage: Move values transactionally

abhinav:
Tracked branch rename reconstructs a stored branch under its new key before
deleting the old key. That obscures whether storage can relocate an existing
value atomically and requires the caller to reproduce the stored record.

Let one update move an optional source before applying sets and deletes.
Use that operation for an unmodified tracked branch record; an earlier change
in the same transaction is written under the new name instead. An absent
source removes a stale destination, and Git retries resolve the source from
the winning revision.

### DIFF
--- a/internal/spice/branch_mutation.go
+++ b/internal/spice/branch_mutation.go
@@ -2,7 +2,6 @@ package spice
 
 import (
 	"context"
-	"encoding/json/jsontext"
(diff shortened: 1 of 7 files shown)


abhinav:

- #1434
    - #1435
        - #1436 ◀

PR #1448, opened 2026-09-21
bitbucket: render nav references as inline cards

krzsztf:
Bitbucket Cloud resolves a link tagged with data-inline-card into the pull request's title and state when the comment is rendered, so a stack navigation entry no longer reads as a bare number. Data Center's CommonMark renderer has no attr_list, so it keeps the plain link.


abhinav (approved):

abhinav:
Thanks!

